PR #3, opened 2024-04-18
CODEOWNERS: add myself, and fix path

zimbatm:


infinisil (approved):
I'll probably soon experiment with https://github.com/marketplace/actions/github-codeowners-validator which could avoid such problems. 

Alternatively I experimented with putting this information into each file individually: https://github.com/nix-open-org/org-v2/commit/a9717a1a78e365fe4a1aaa33ea2c67e6fbba11f8. This would be easier to keep track of, though it makes automation harder since GitHub doesn't automatically request reviews like that. Should be doable though.

zimbatm:
Something I haven't thought through, but codeowners can only include people who have commit access to the repo.

PR #4, opened 2024-04-23
Validate CODEOWNERS

infinisil:
This starts validating the CODEOWNERS file using https://github.com/mszostok/codeowners-validator and, because that project [doesn't support it yet](https://github.com/mszostok/codeowners-validator/issues/157), a custom script to make sure that everybody has write access.

To make this work, I introduced the `OWNERS_VALIDATOR_GITHUB_SECRET` secret to contain a read-only personal access token of mine for now (expires in 30 days). This is needed because it's not possible to query for who has write access without that. In the future we should [switch to a GitHub App](https://github.com/mszostok/codeowners-validator/blob/main/docs/gh-auth.md#github-app) instead.

Because we need to switch to `pull_request_target` to trigger CI, I had to test this in a separate PR, but it works:
- [Adding a new file without a code owner](https://github.com/nix-open-org/org-v2/actions/runs/8807634626/job/24175135285?pr=5)
- [Adding a code owner without write access](https://github.com/nix-open-org/org-v2/actions/runs/8807870347/job/24175876763?pr=5)

Furthermore, the check to make sure everybody has write access is also run in the regular monthly review, which I can confirm to work correctly:
- https://github.com/infinisil/org-v2/issues/5


infinisil:
Now also added @zimbatm as codeowner to all files that didn't have one before.

zimbatm (approved):

zimbatm:
Thanks!

PR #7, opened 2024-04-25
Make new org team codeowner

infinisil:
Over time we should assign others, but for now this makes the most sense

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,23 +1,23 @@
 # See doc/org-repo.md
 
-/README.md @infinisil @zimbatm
-/CONTRIBUTING.md @infinisil @zimbatm
-/LICENSE @infinisil @zimbatm
+/README.md @NixOS/org
+/CONTRIBUTING.md @NixOS/org
+/LICENSE @NixOS/org
 
-/.gitignore @infinisil @zimbatm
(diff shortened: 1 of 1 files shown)


infinisil:
This is not a problem with this PR itself, but it's a good time to address it:

We wanted to switch to [GitHub App authentication](https://github.com/mszostok/codeowners-validator/blob/main/docs/gh-auth.md#github-app) for the codeowners validation, which would work for https://github.com/mszostok/codeowners-validator, but not the custom script I wrote to check that everybody has write access (which I wrote because it's [not supported by codeowners-validator](https://github.com/mszostok/codeowners-validator/issues/157)), since the script relies on a GitHub token and it's non-trivial to authenticate as a GitHub App instead.

Well, instead of trying to hack around with more bash, I decided to implement this feature in codeowners-validator instead: https://github.com/mszostok/codeowners-validator/pull/222 :smile:

zimbatm (approved):
:LGTM:

zimbatm:
I think this will have to be force-merged due to the target_pull_request rule

infinisil:
I'm not making use of my fix yet, I'll try that soon. Let's not merge until then :)

Edit:
- https://github.com/NixOS/org/pull/8

infinisil:
Now that #8 is merged, this passes CI and is ready to be merged :)

PR #8, opened 2024-04-26
Use GitHub App for codeowner validation and remove hacky script

infinisil:
We shouldn't use personal access tokens, instead we created a GitHub App with read-only access to just this repository.

While codeowners-validator supports GitHub App authentication, the same cannot be said for the hacky script I wrote because there was no support for checking write access: https://github.com/mszostok/codeowners-validator/issues/157

Instead of trying to hack the script more to make it work with GitHub App authentication, I decided to implement it into codeowners-validator itself: https://github.com/mszostok/codeowners-validator/pull/222

Because it's not merged/released yet, we need to build it ourselves, so I added some Nix to do that reproducibly.

Testing this in https://github.com/NixOS/org/pull/9


infinisil:
We can't make CI pass here because it uses the base branch's workflow file, which doesn't use GitHub Apps yet.

However I tested this in a separate PR where I can confirm that it works: https://github.com/NixOS/org/pull/9

zimbatm (approved):

PR #57, opened 2025-01-16
Require org owners to use exclusively secure two-factor authentication

infinisil:
Turns out @tomberek and @Lassulus [don't right now](https://github.com/orgs/NixOS/people?query=role%3Aowner+two-factor%3Ainsecure), I think we should.

### DIFF
--- a/doc/github-org-owners.md
+++ b/doc/github-org-owners.md
@@ -7,6 +7,7 @@ The following people have the GitHub "owners" permissions on the NixOS organizat
 - [@tomberek](https://github.com/tomberek)
 - [@winterqt](https://github.com/winterqt)
 - [@zimbatm](https://github.com/zimbatm)
+
 ## How to contact the team
 For any GitHub-related needs, you can reach out to the org owners by either:
 - Pinging [@NixOS/org](https://github.com/orgs/NixOS/teams/org)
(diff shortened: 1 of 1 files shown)


winterqt (approved):
1000000%.

(Isn't the only other option SMS?)

Gabriella439 (approved):
I don't have a strong opinion on this, but it seems like a good idea

jtojnar (approved):
Sounds reasonable.

jtojnar:
Ideally, we would have GitHub enforce it but, unfortunately, it looks like it is not possible to force require it just for owners.

It might be a good idea to eventually require it for everyone but not sure how many people it would affect as https://github.com/orgs/NixOS/people?query=two-factor%3Ainsecure shows the same list for me as https://github.com/orgs/NixOS/people?query=two-factor%3Asecure

infinisil:
@jtojnar I don't think that needs to be discussed here, but it shows two different lists for me: 66 pages for secure and 57 pages for insecure, so secure 2FA seems to be at least a close majority already.

Lassulus:
not sure what I need to do that github deems my 2FA as secure :D

infinisil:
@Lassulus As Winter said, I think SMS is the only method not considered secure, can you confirm that you have that enabled ([direct link to account security settings](https://github.com/settings/security))? If so, you'll have to turn that off

fpletz (approved):

Ericson2314 (approved):

tomberek:
I've deleted SMS auth. Does that resolve the insecure issue?

infinisil:
Yes [it does](https://github.com/orgs/NixOS/people?query=role%3Aowner+two-factor%3Ainsecure), thanks! I'll merge this as we have SC majority and it matches reality now :)

PR #44, opened 2024-12-02
clarify discourse administration

Lassulus:
roles are now clarified

PR #58, opened 2025-01-20
Add methods for contacting entire SC

infinisil:
So far the email was only announced [here](https://discourse.nixos.org/t/nix-steering-committee-kickoff/56362) and the GitHub team was never mentioned.

### DIFF
--- a/doc/governance.md
+++ b/doc/governance.md
@@ -13,6 +13,8 @@ The current SC consists of these members:
 - [@tomberek](https://github.com/tomberek) (term ends 2025)
 - [@winterqt](https://github.com/winterqt) (term ends 2025)
 
+The entire SC can be reached via <steering@nixos.org> or by pinging the [@NixOS/steering](https://github.com/orgs/nixos/teams/steering) GitHub team.
+
 ## RFC Process
 
 The Nix RFC (Request For Comments) process exists to establish wider consensus for substantial changes. See the [rfcs repository](https://github.com/NixOS/rfcs) for more information.


jtojnar (approved):

fricklerhandwerk (approved):

Gabriella439 (approved):

fpletz (approved):

doc/governance.md
```diff
@@ -13,6 +13,8 @@ The current SC consists of these members:
 - [@tomberek](https://github.com/tomberek) (term ends 2025)
 - [@winterqt](https://github.com/winterqt) (term ends 2025)
 
+The entire SC can be reached via <steering@nixos.org> or by pinging the [@NixOS/steering](https://github.com/orgs/nixos/teams/steering) GitHub team.
```
winterqt, line 16: There's also the [Discourse group](https://discourse.nixos.org/g/steering_committee), which AFAIK can be directly PM'd. Worth adding?

mweinelt:
I did ask for a Matrix room before, because that is the official chat platform we have. 

And on Matrix @tomberek agreed it would be good to have one: https://matrix.to/#/!CJXQiUGqNPcFonEdME:nixos.org/$RkbRsc-ILYB6YtCI5GwsjJPFnEXdiOZx2HMj12w81K0?via=nixos.org&via=matrix.org&via=nixos.dev

doc/governance.md
infinisil: It appears to not be possible, autocompletion doesn't bring up the SC. I believe the Discourse admins (#44 !) need to configure something for that to work. Notably the moderators can be PMd, but it autocompletes as "moderators", not ["moderation-team" which would be a Discourse group](https://discourse.nixos.org/g/moderation-team), which leads me to believe it's something different than Discourse groups that makes PMs possible.

mweinelt: Whether or not a group can be mentioned and by whom is a per group setting. Right now nobody can mention the steering committee.

fpletz:
> I did ask for a Matrix room before, because that is the official chat platform we have.

I'm also not opposed to the idea. Just wondering if a more general "org" channel for the foundation, SC and other org-wide committees may make more sense since I would presume it's very low traffic anyway and most people doing org stuff are interested in all the discussions anyway. If there are less channels in total with each having with a clear topic, I think more people would join.

tomberek (approved):

infinisil:
For now this seems good to merge, we can use follow-up PRs for additional means of contact.

PR #60, opened 2025-01-22
Expand org owner authority

infinisil:
I'd like to be able to remove the security and systemd team's explicit Nixpkgs write access, which seems to be a historical artifact from needing to have write access to be requested for reviews with CODEOWNERS.

The current authority doesn't allow for that, so this PR is expanding it to include that and similar changes. Notably the expansion is conditioned under non-controversial changes, and needing approval from two org owners.

### DIFF
--- a/doc/github-org-owners.md
+++ b/doc/github-org-owners.md
@@ -29,10 +29,9 @@ All org owners can individually take care of implementing:
 
 Org owners need approval from at least one other org owner to take care of implementing
 higher-impact changes that are _not controversial_, such as:
-- Administer unmaintained repos, such as:
-  - Performing maintenance.
-  - Giving commit access to trusted people that offer maintenance.
-  - Archiving if appropriate.
+- Changes to permissions of teams and people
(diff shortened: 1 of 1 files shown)


doc/github-org-owners.md
```diff
@@ -29,10 +29,9 @@ All org owners can individually take care of implementing:
 
 Org owners need approval from at least one other org owner to take care of implementing
 higher-impact changes that are _not controversial_, such as:
-- Administer unmaintained repos, such as:
-  - Performing maintenance.
-  - Giving commit access to trusted people that offer maintenance.
-  - Archiving if appropriate.
+- Changes to permissions of teams and people
```
winterqt, line 32: If we want this to be *non-controversial*, perhaps this should be worded as something like "with the team's approval?" Unless that's implied.

winterqt (approved):

infinisil:
Not a fan of having to list everything explicitly, I'll make another PR.

infinisil:
I opened https://github.com/NixOS/org/pull/61 instead.

PR #61, opened 2025-01-22
Clarify and extend org owner authority

infinisil:
I've been annoying the other org members with my complaints when we didn't follow the authority to the dot :sweat_smile:

This was because:
- The "such as" wording left it open to ambiguity as to whether it's only the listed items or also others (while proper English would imply the latter, I meant the former when I wrote this originally)
- The former interpretation is unnecessarily restrictive, requiring further requests to increase authority whenever something new was being done.

This PR makes it very explicit that it's an exclusive list for individually-implementable changes (certain pre-approved changes and minimal-impact changes), while leaving _all_ other non-controversial and non-impactful changes up to the org owners, given two approvals.

This should hopefully require less changes to this document in the future.


Gabriella439 (approved):

tomberek (approved):

wolfgangwalther (approved):

infinisil:
Closing for now, shouldn't leave this lingering for so long, might need to rethink

PR #66, opened 2025-01-31
Document vault with owners and admins

infinisil:
I [asked about this on Matrix](https://matrix.to/#/!RROtHmAaQIkiJzJZZE:nixos.org/$VJTHDLv9dCiHSBJd_G5H5RLFD6uidISeS8Kh3-haQaU?via=nixos.org&via=matrix.org&via=nixos.dev) recently. There's more that could be written down (or automated), but this is a good first step.

@zimbatm <s>@mweinelt</s> @Mic92: Please confirm that the written docs are still accurate

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -22,6 +22,7 @@
 /doc/resources.md @NixOS/steering
 /doc/governance.md @NixOS/steering
 /doc/values.md @NixOS/steering
+/doc/vault.md @Mic92 @zimbatm
 /doc/constitution.md @NixOS/steering
 /doc/nixpkgs-committers.md @Mic92 @NickCao @jtojnar
 /doc/calendar.md @edolstra @tomberek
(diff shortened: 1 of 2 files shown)


zimbatm (approved):
LGTM

mweinelt:
image

infinisil:
Ahh whoops, will fix

infinisil:
Done

mweinelt:
I am not sure why the roles are as they are, but I am as surprised as you are.

infinisil:
It's consistent with what you [said before](https://matrix.to/#/!RROtHmAaQIkiJzJZZE:nixos.org/$EfFGsdAgvMpKjLyucg-gBKH8hQhxM3Um2C0kanTq4bE?via=nixos.org&via=matrix.org&via=nixos.dev), I just mixed it up at first, so I'm not surprised 😄

Why are you surprised though?

zimbatm:
It doesn't make sense why Mic92 would be an owner and not Hexa. Documenting is the first step.

Mic92:
Here is the difference between these roles: https://bitwarden.com/help/user-types-access-control/

Don't really see the advantage of having admins vs owner. We can just making all admins owner for simplicity sake.

infinisil:
https://github.com/NixOS/org/pull/69

PR #72, opened 2025-02-16
Enable SC group messages on Discourse

infinisil:
Doesn't work right now, needs to be implemented before merged, probably by @jtojnar as the SC-representative Discourse admin.

This was [discussed before](https://github.com/NixOS/org/pull/58#discussion_r1924177806).


jtojnar (approved):
Enabled it:

Discourse settings

infinisil:
@jtojnar The convention for this repo is to always merge and implement at the same time and ideally by the same person, so feel free to merge too ;)

Issue #70, opened 2025-02-10
NixCon 2025


infinisil:
Planning for NixCon 2025 seems to have been initiated, let's have this issue to discuss it in general (hopefully later turned into a PR).

infinisil:
Here's what I know:

- @Lassulus seems to be initiating the organisation
- The [#nixcon2025-orga:lassul.us Matrix room](https://matrix.to/#/#nixcon2025-orga:lassul.us) is used for discussing organisation
- @Lassulus is proposing to hold it in Switzerland at the venue proposed by @ners for 2024: https://github.com/nixcon/nixcon-proposals/issues/4. The local crew was asked about that [in the Nix Zürich Matrix room](https://matrix.to/#/!oeFELgatfbdiAEkJAY:nixos.org/$UpmMWfUjsYSb2EV85dkb9UnznMPDE6hrAOo8mbpSqGI?via=nixos.org&via=matrix.org&via=nixos.dev).
  - @ners is happy to help with organisation
- Budget is a concern, there's a budget plan from last year: https://docs.google.com/spreadsheets/d/1P2ddDFcOpUStXN2mjBLlVgyVKQFxex75InaHgXiZVQw/edit?usp=sharing
- @zmberber is also happy to help, an early decision for the location would be great

This makes me ask the question: Who makes the final decisions for NixCon? It's de-facto official, but not yet properly documented without #22. If official but ownership is unclear, it could be up to the @NixOS/steering to either decide or delegate.

Also ccing @zimbatm, @andir, @john-rodewald, @das-g

zimbatm:
The way NixCon *always* happened is that somebody decides to step up and organize it this year.

They find a location, assemble a motivated crew, find their own budget (eg: sponsors and tickets), and let's go. For most years, we had to bring our own fiscal host, but since 2023 the foundation can take care of that part.

Then all Andi or I do is make sure they have access to the domain, github, pretix, pretalx, mailing list, etc.

In 2024 we tried adding something more stuctured with the [NixCon PL team](https://discourse.nixos.org/t/governance-through-project-leads-in-nixcon-2024/33981), but all it did at the end was delay and complicate things. We almost didn't have a NixCon that year because of that. Then @Lassulus and his crew stepped up and threw one of the best NixCon ever. To me this is one more comfirmation that we should empower the people that want to organize it, and stay out of the way otherwise.

[...]

PR #74, opened 2025-03-03
NixCon is official, 2025 organisation and location (OST in Rapperswil-Jona, Switzerland)

### LINKED ISSUE #22: Recognise NixCon as a public communication resource
Not sure how exactly it should be worded, but NixCon is probably treated as official

infinisil:
As mentioned in #70, we had a first meeting today to start organisation of NixCon 2025. Present were @zmberber, @a-kenji, @lassulus, @pinpox, @refroni, @Nebucatnetzer and @infinisil. While @Nebucatnetzer and @infinisil are local, the more experienced local event organisers @ners, @john-rodewald and @das-g could not attend this first meeting.

In the meeting we agreed to now work with the assumption that NixCon 2025 _will_ be held at [OST Rapperswil-Jona (Switzerland)](https://github.com/nixcon/nixcon-proposals/issues/4#event-13200457377). While there are many things to figure out still, **if we can make OST Rapperswil-Jona happen, let's make it happen**.

Furthermore, we generally agreed that:
- NixCon should be made official and we should be delegated explicitly as organisers for this year.
- If we cannot reach consensus on important decisions, they are escalated to the Steering Committee.
- Decisions relating to sponsorships are always escalated to the Steering Committee.

This PR is still a draft because we should get confirmation from @ners, @john-rodewald and @das-g for the location and for the vital local organisation.

This needs to be approved by the SC. To implement this PR, the following also needs to happen:
- The owners of the both [NixOS](https://github.com/nixos) and [NixCon](https://github.com/nixcon) GitHub organisations should be made to match (@zimbatm or @lassulus can implement this)
- The <nixcon@nixos.org> email should be aliased to the organisers (infra team can implement this, @infinisil has the list of emails). Related: #75 
- The https://github.com/nixcon/2025.nixcon.org/ repository should be created, the organisators should be given write access to it

This PR closes https://github.com/NixOS/org/issues/22

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -26,3 +26,4 @@
 /doc/constitution.md @NixOS/steering
 /doc/nixpkgs-committers.md @Mic92 @NickCao @jtojnar
 /doc/calendar.md @edolstra @tomberek @fricklerhandwerk
+/doc/nixcon.md @NixOS/steering
--- a/doc/github-org-owners.md
+++ b/doc/github-org-owners.md
@@ -1,6 +1,8 @@
 ## GitHub org owners
 
-The following people have the GitHub "owners" permissions on the NixOS organization:
+Both the [NixOS](https://github.com/nixos) and [NixCon](https://github.com/nixcon) GitHub organisations are official.
+
+The following people have the GitHub "owners" permissions:
 <!-- Also keep this in sync with the members of @NixOS/org! -->
 - [@infinisil](https://github.com/infinisil)
 - [@lassulus](https://github.com/lassulus)
--- a/doc/nixcon.md
+++ b/doc/nixcon.md
@@ -1,0 +1,14 @@
+# NixCon
+
+NixCon is a yearly official Nix conference by and for the community.
+The legal and fiscal backing is provided by the NixOS Foundation.
+The conference has previously been held in these locations:
+
+- 2015: Berlin (Germany), http://2015.nixcon.org/
+- 2017: Unterföhring (Germany) http://nixcon2017.org/
+- 2018: London (UK), https://2018.nixcon.org/
+- 2019: Brno (Czech Republic), http://2019.nixcon.org/
+- 2020: (online), http://2020.nixcon.org/
+- 2022: Paris (France), https://2022.nixcon.org/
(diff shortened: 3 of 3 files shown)


infinisil:
There was another meeting today with @fmehta, @ners, @john-rodewald (Edit: @das-g couldn't make it) (and again with @Nebucatnetzer, @hrmny, @Lassulus and @ra33it0), with everybody on board to make it happen at OST. I also added @andir and @ra33it0 to the organisers.

We now need approval from @NixOS/steering to continue with planning, the sooner the better!

Gabriella439 (approved):

fpletz (approved):
My only concern here is that the prices in Switzerland are quite high. That might prevent some people from coming. I would encourage to try to organize some kind of cheap shared accommodation for people that need it. No hard blocker though. Great to see a team stepping up to organize. :+1:

Ericson2314 (approved):

Nebucatnetzer:
> My only concern here is that the prices in Switzerland are quite high. That might prevent some people from coming. I would encourage to try to organize some kind of cheap shared accommodation for people that need it. No hard blocker though. Great to see a team stepping up to organize. 👍

Yeah there are ideas floating around to make it cheaper and therefore more accessible for everyone.

tomberek (approved):

jtojnar (approved):

infinisil:
(the merger ideally implements the changes themselves, but in this case no single person can do that, so instead it's better to think of the merger as "the person _responsible_ for implementing the changes")

I've opened #76 to set up the email alias, for the infra team to implement

And I'll ask @zimbatm (or @Lassulus as a fallback) to update the org owners of the NixCon org to match the NixOS org ones

zimbatm:
Synched the NixCon org owners with the NixOS ones. You should have received an invite.

infinisil:
Now that I'm also org owner of the NixCon GitHub org, I've created a basic https://github.com/nixcon/2025.nixcon.org mostly copied from 2024, gave the organisers write access, and set it up to deploy to the default https://nixcon.github.io/2025.nixcon.org/ (we can set up 2025.nixcon.org and nixcon.org later)

PR #83, opened 2025-03-14
Update nixcon.md

idabzo:
added hrmny and ida to the team


infinisil (approved):
I don't think we need full SC approval for this addition, since being an organiser doesn't come with much authority or responsibility by default, instead it's just to keep track of who's interested in helping out, which I can confirm :D

Still, we need one SC member to approve because of code ownership.

Btw we'll also need to update the nixcon email alias to include the new users, can be in this PR or another one.

jtojnar (approved):

Issue #40, opened 2024-11-30
RFC: Manage the org programatically


zimbatm:
A while back, @infinisil and I considered managing the GitHub Organization as code. To get something out quickly, we decided to start with today's text-only format.

The vision for the code approach was that any members could submit PRs to change their team membership instead of asking us to do the changes manually. Assuming CI is applying the changes, this would allow the SC to delegate ownership of the org more effectively while maintaining a level of transparency since all the changes are happening through code now. The "org" team can still be around as a fallback but doesn't need to be invoked as often.

A few months back, I met with the wonderful people behind https://www.ipdx.co/ (@galargh and @laurentsenta), and they have a ready-made solution: https://github.com/ipdxco/github-as-code . It works both ways: the data model updates the GitHub org settings, and the org setting changes also update the data model. It's based on Terraform.

Before doing anything, I'd like to get the feedback of the SC and other org owners on whether that would be a good idea. Let me know if I should get into more details.

In terms of migration paths, I see something like this:
1. Configure a new Terraform backend in AWS with the Infra team.
2. Fork https://github.com/ipdxco/github-as-code as an org-2 repo.
3. Setup CI.
4. Leave the nixpkgs-maintainers team out of this for now, since it is managed by a separate bot.
5. Party!

galargh:
@zimbatm Thanks for putting together the issue! I'd be more than happy to help with the setup or any other changes to the GitHub as Code project if needed 😄

Issue #67, opened 2025-02-02
Github app for merge-bot and for nixos-infra  for updating flakes


Mic92:
I need two github apps for both merge-bot and nixos-infra,
so that I can update create pull requests in a way that trigger CI,
Similar to what we do for nixpkgs: https://github.com/NixOS/nixpkgs/blob/c1f95f184125fe9dbeed9f2fb39596c8f356502c/.github/workflows/backport.yml#L26C1-L26C65

See https://github.com/NixOS/org/issues/38 for a tutorial how to do set this up.

If you at it, I would like one app for the Nix repository as well. Than we can use github builtin merge queues instead of mergify.

infinisil:
How about we repurpose the Nixpkgs CI app for other repos too? There's no reason we'd need per-repository apps, it just makes it more annoying to manage. So we could:
- Rename it to something more generic like "Nix Projects CI"
- Install it on _all_ repos instead of individual repos

Then the only per-repository thing we need to do is to generate a private key and set the repository variables/secrets.

infinisil:
And at that point we could go even further and specify organisation-wide GHA variables/secrets, such that _all_ repos can get access to the GitHub app.

If we do this though, it also means that GHA's on any repo could write to any other repo, but I think that's fine because you still need somebody with commit access to configure GHA's.

What do you think, @NixOS/org?

zimbatm:
Whenever I give commit access to somebody, I can reason that their area of action is within the repo. I would prefer to keep that property and avoid unintended consequences down the line. Especially if it becomes easier to make more projects official.

Mic92:
I also want to keep these things separate. Otherwise every nixpkgs committer for example could commit to the nix repository.

infinisil:
Fair yeah. Especially also because there are a bunch of repos that aren't watched by many, and ao could more easily be sneaked in some changes.

infinisil:
Before I go ahead with this, here's the plan:

Two GitHub Apps:
- One installed on https://github.com/nixos/nixpkgs-merge-bot named "Nixpkgs Merge Bot CI"
- One installed on https://github.com/nixos/infra named "Nix Infra CI"

Each with only these permissions on the respective repository:
- Repository > Contents: read-write (to create a PR branches)
- Repository > Pull Requests: read-write (to create PRs)

Each will be available via:
- `${{ vars.CI_APP_ID }}`
- `${{ secrets.CI_APP_PRIVATE_KEY }}`

[...]

PR #99, opened 2025-04-10
Introduce github-mgmt repository

infinisil:
After me and @winterqt have [experimented](https://github.com/Infinisil-s-Test-Organization/github-as-code) with the suggestion in NixOS/org#40 of using [github-as-code](https://github.com/ipdxco/github-as-code), we're confident that we should deploy for real, so that people can more easily request GitHub org changes by just making PRs.

As this is strictly outside the [authority of the org owners](https://github.com/NixOS/org/blob/main/doc/github-org-owners.md#authority-and-processes), I am hereby formally requesting @NixOS/steering approval to **give the org owners authority over the deployment of github-as-code**.

Some notes:
- github-as-code _bi-directionally_ syncs changes, so any imperative changes to the GitHub org will get synced back to the repo automatically, it will not interfere with that. But in addition, merged PRs will get synced to the GitHub org too.
- Only the org owners will be able to make changes to the repo for a start, but this could be expanded in the future with good file separation and codeowner declarations like the org repo.
- We'll only use it for team memberships to start out
  - Should have an exception for the Nixpkgs maintainers team, depends on https://github.com/ipdxco/github-as-code/pull/162
- This will pave the way towards https://github.com/NixOS/nixpkgs-committers/issues/62 (which I'd also still like SC approval on)

### DIFF
--- a/doc/github.md
+++ b/doc/github.md
@@ -64,6 +64,12 @@ The home of the branding guide and media kit.
 
 The [marketing team](https://nixos.org/community/teams/marketing/) controls this repository.
 
+### [github-mgmt](https://github.com/NixOS/github-mgmt)
+
+Bi-directionally syncs a subset of the GitHub org settings.
+
+The [GitHub org owners](./github-org-owners.md) own this repository.
+
 ## Groups
 
 ### [nixpkgs-committers](https://github.com/orgs/NixOS/teams/nixpkgs-committers)


fricklerhandwerk:
This would also finally open the doors to automatically list team members on nixos.org (ping @avocadoom)!

Gabriella439 (approved):

winterqt (approved):

jtojnar (approved):

roberth (approved):
Approving conditional on testing against
- accidental deletions (e.g. can a repo and all its metadata be deleted or what would it take?)
- a plan B for the possibility that all admins that are managed by this system are locked out

tomberek:
As another data point: https://github.com/apache/infrastructure-asfyaml

infinisil:
Thanks for the SC approval! We (org owners) will evaluate this more before committing because there's some security concerns:
- The github-as-code App needs a lot of permissions and GitHub doesn't allow narrowing it down. This most notably includes:
  - The [Organization Members permission](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps#organization-permissions-for-members), which allows both management of teams, but also removal of admins
  - The [Repository Administration permission](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-administration), which allows both management of repo settings, but also removal of repos
  - The [Repository Contents permission](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents), which allows pushing commits to any repo (if branch protection allows it), which is used to implement the bi-directional syncing
- github-as-code regularly runs code from the very same repo that it also pushes bi-directional syncing commits to -> self-modifying code isn't that far off
- The standard way of upgrading github-as-code is to manually trigger the [upgrade action](https://github.com/ipdxco/github-as-code/blob/master/.github/workflows/upgrade.yml), which syncs the latest changes from upstream and [pushes it directly](https://github.com/ipdxco/github-as-code/blob/9f205a7d69e8e68909f3f0e710aa7a9100e46a8e/.github/workflows/upgrade_reusable.yml#L58) without a PR

We can do:
- (easy) Reach out to contacts at GitHub to get their thoughts on this (@tomberek is on this)
- (easy) Test whether admins can be locked out as suggested (I suspect it's definitely possible)
- (medium) Review upstream for which API calls are used, both initially and continually before upgrading in the future
  - (easy) Make the upgrade action create a PR instead
  - (medium) Fork github-as-code and manually pull in updates
- (medium) Ensure that we have backups of everything that github-as-code could wipe in theory (could be fairly easy since it itself writes all of the state to an S3 bucket, whose history we can preserve)
- (medium) Ask github-as-code to onboard trusted members from our side as maintainers
- (medium) Remove the need for the global Repository Contents permission by separating that permission to another GitHub App scoped to just the github-mgmt repository.
- (hard) Properly separate code from config by having a separate repository for them. This would be hard to upstream.
- (hard) Find or build a GitHub-imitating API server that only allows exactly the API calls that we need.
- (hard) Ask GitHub to add another permission scope for GitHub Apps needed for potentially destructive operations like removing admins and repos.

With some of the above I expect us to have enough confidence in the deployment, but we'll see

tomberek:
Another data point: https://github.com/eclipse-csi/otterdog

winterqt:
Some thoughts:

> The [Repository Contents permission](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents), which allows pushing commits to any repo (if branch protection allows it), which is used to implement the bi-directional syncing

For this one specifically, I find it a bit odd that we can't restrict it -- realistically, we only need to give the app content access to the management repo, since we aren't using its capability to add files to other repos in the org.

> (easy) Test whether admins can be locked out as suggested (I suspect it's definitely possible)

Couldn't we give the app member permissions, but not make it an owner itself, which *may* allow some security in this sense (that is, it can't remove non-owners)?

> - (medium) Review upstream for which API calls are used, both initially and continually before upgrading in the future
>     - (easy) Make the upgrade action create a PR instead
>     - (medium) Fork github-as-code and manually pull in updates

I like this.

> (medium) Ensure that we have backups of everything that github-as-code could wipe in theory (could be fairly easy since it itself writes all of the state to an S3 bucket, whose history we can preserve)

Doesn't prevent things like issues/wiki/etc from being wiped -- we'd need good backup solutions for those.

> (medium) Ask github-as-code to onboard trusted members from our side as maintainers

We'd have to ask, but this may be doable, esp. given we'll probably be its biggest (in terms of members) user.

> (medium) Remove the need for the global Repository Contents permission by separating that permission to another GitHub App scoped to just the github-mgmt repository.

Ah, yeah, exactly what I said above. :)

infinisil:
> For this one specifically, I find it a bit odd that we can't restrict it -- realistically, we only need to give the app content access to the management repo, since we aren't using its capability to add files to other repos in the org.
> [...]
> > (medium) Remove the need for the global Repository Contents permission by separating that permission to another GitHub App scoped to just the github-mgmt repository.
>
> Ah, yeah, exactly what I said above. :)

Yeah so we need two GitHub Apps for that, because any single GitHub apps permissions are only restricted to essentially the cross product of repos and permissions, so it's not possible to give it certain permissions for one repo and different permissions for another/all.

> Couldn't we give the app member permissions, but not make it an owner itself, which may allow some security in this sense (that is, it can't remove non-owners)?

Permissions of GitHub apps are separate from the user model, the App isn't an owner itself. That said, it's a really interesting idea to make it use the user model instead:
- Create a dedicated user
- Give it the team maintainer permission on all teams
- Give it maintainer permission on all repos, [described as](https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization) "Recommended for project managers who need to manage the repository without access to sensitive or destructive actions"
- Give it write access to only the `github-mgmt` repo.
- Create a personal access token for the user and make github-as-code use it instead of the App

This way there would be no way for the tool to remove admins nor repos! Frankly this sounds like a really good option, probably medium difficulty since it requires some upstream changes.

---

Separately I'd also like to highlight @arianvp's [suggestion here](https://github.com/NixOS/org/issues/67#issuecomment-2827537059) and the associated discussion as a potential alternative security model.

infinisil:
I've submitted https://github.com/ipdxco/github-as-code/pull/198 to implement the above-mentioned idea, and confirmed that it works for team management!

wolfgangwalther:
> After [we] have [experimented](https://github.com/Infinisil-s-Test-Organization/github-as-code) with the suggestion in NixOS/org#40 of using [github-as-code](https://github.com/ipdxco/github-as-code),

I looked into that briefly - am I correct that this is "only" about managing teams and memberships? Or can/will repo settings be managed this way as well, for example branch protection rules?

Especially branch protection rules are potentially complicated, so would be very nice to have them transparently available as code.

> Give it maintainer permission on all repos, [described as](https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization) "Recommended for project managers who need to manage the repository without access to sensitive or destructive actions"

Seems like only admins can configure branch protection rules, but not maintainers. So this approach would rule out the above.

wolfgangwalther:
It seems like github-as-code only supports the old branch protection rules and not the newer rulesets, which are a lot more flexible. At least that's what the example in the repo tells me: https://github.com/Infinisil-s-Test-Organization/github-as-code/blob/master/docs/EXAMPLE.yml. Since the branch pattern to match on is the yaml key, it seems very unlikely that this can support the more advanced `exclude` and `include` pattern of rulesets.

And since that's 2-way sync, I assume those rulesets would also not show up in the repo, when we configure them in the settings... it looks to me like github-as-code might be too limited.

infinisil:
> I looked into that briefly - am I correct that this is "only" about managing teams and memberships? Or can/will repo settings be managed this way as well, for example branch protection rules?

github-as-code supports repo settings as well, but for now we're only interested in managing team with it.

I don't think it would be a big issue for github-as-code to support rulesets as well (in theory), YAML keys can just be quoted, and the DSL of the yaml can be arbitrary.

The nice thing about rulesets though is that they are transparent by default. E.g. check out the [branches](https://github.com/NixOS/nixfmt/branches) in the nixfmt repo and click on the little :shield: icon next to `master`, which lets you explore the rulesets.

infinisil:
Closing this for now, needs more thought

PR #100, opened 2025-04-16
fix: hrmny matrix username

ForsakenHarmony:


infinisil:
I've just opened https://github.com/NixOS/org/pull/101 as an alternative, proposing to use https://github.com/nixcon/2025.nixcon.org/blob/main/src/pages/Organizers.tsx as the source of truth for NixCon organisers. Notably all organisers have write access to that repo, so @ForsakenHarmony you can push the change to your Matrix handle yourself :D

winterqt:
Superseded by #101.

infinisil:
Fixed the handle in https://github.com/nixcon/2025.nixcon.org/commit/de5d471f065d9dd7cf2d9bdefd93c26299f20625

PR #101, opened 2025-04-16
Track NixCon organizers externally

infinisil:
Especially now that the website is listing them.
Also good because the SC shouldn't have to approve changes to that


Gabriella439 (approved):

infinisil:
Whoops: https://github.com/NixOS/org/pull/102

PR #108, opened 2025-05-11
Committers process update

infinisil:
This ensures that all commit delegators can always be comfortable with all committers.

CC @winterqt 

Note that as this changes the commit delegation process, it needs approval from all commit delegators.

### DIFF
--- a/doc/nixpkgs-committers.md
+++ b/doc/nixpkgs-committers.md
@@ -18,4 +18,6 @@ They can be contacted either individually, or through a [group message on Discou
 ## Process
 - The process must be publicly documented (this document)
 - Any change to the process must be unanimously agreed upon by all delegation team members
-- Any change in the list of committers must be publicly documented and agreed upon by all delegation team members
+- Any change in the list of committers must be publicly documented
+- Addition of a committer must be agreed upon by all delegation team members
+- Removal of a committer only requires agreement of a single delegation team member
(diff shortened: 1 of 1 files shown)


NickCao (approved):
Reasonable!

zimbatm (approved):

infinisil:
(also needs approval from @jtojnar and @winterqt now)

doc/nixpkgs-committers.md
```diff
@@ -18,4 +18,6 @@ They can be contacted either individually, or through a [group message on Discou
 ## Process
 - The process must be publicly documented (this document)
 - Any change to the process must be unanimously agreed upon by all delegation team members
-- Any change in the list of committers must be publicly documented and agreed upon by all delegation team members
+- Any change in the list of committers must be publicly documented
+- Addition of a committer must be agreed upon by all delegation team members
+- Removal of a committer only requires agreement of a single delegation team member
```
winterqt, line 23: While I like this, I fear it gives any one individual in the team too much power -- which is maybe a good thing! -- but perhaps something we should discuss the impact of in more detail before approving.

winterqt (approved):
We've discussed this internally, I think we can give it a go.

jtojnar (approved):
Yeah, could potentially lead to instability but if the committer delegation team cannot maintain accord internally, we have a bigger problem.

PR #115, opened 2025-05-28
nixcon: Tune the sponsorship approval process

infinisil:
Before, there was the implied requirement to wait on organiser objections for 1 week, with no way to skip if the sponsor is uncontroversial.

This commit refines it to be clear that organisers should issue objections within 2 days, and only if any objections are issued within that time, the SC needs to reevaluate whether the sponsor can be approved.

This allows speeding up the sponsor approval process.

This change is somewhat weakening the NixCon organisers power over sponsorships, I'll collect organiser feedback on this tomorrow. If there's no objections, I think this can go ahead with just a single SC approval.

### DIFF
--- a/doc/nixcon.md
+++ b/doc/nixcon.md
@@ -94,8 +94,8 @@ The following is internal documentation and should not go on the website. For 20
 1. After receiving the email, board respondent replies that the email was received and that they will hear back within about 2 weeks.
 2. SC respondent ensures the SC internally rejects or provisionally approves the sponsor within 1 week.
    - If rejected, SC respondent replies to the sponsor with the rejection message.
-3. If provisionally approved by the SC, SC respondent forwards the name and website of the sponsor to the NixCon orga Matrix room for organiser approval.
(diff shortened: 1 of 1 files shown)


winterqt (approved):
Reasonable.

infinisil:
Just so it doesn't get merged too early: This is still waiting for some looks from the other organisers

Nebucatnetzer:
Makes sense to me.

das-g:
Two days seem a very short to me, as I sometimes have more than 48h of messages to catch up on. Then again, for any potential sponsor I'd have objections about, it's unlikely that I'd be the only one of the organizers to object, so chances would be high that some other organizer would raise the objection in time (if such a potential sponsor even got the provisional SC approval at all.)

ners (approved):

Lassulus (approved):

john-rodewald (approved):

tomberek (approved):

Gabriella439 (approved):

roberth (approved):

Issue #120, opened 2025-06-09
Use new "rulesets" for branch protection instead of "classic branch protections" in all repos


wolfgangwalther:
There are two kinds of branch protection configuration we can use:
- The classic branch protection under "Branches" in the menu
- The new rulesets under "Rules / Rulesets"

Currently, nixpkgs and probably most repos in the org, are using the classic branch protection rules. However, the new rulesets have multiple advantages:
- They are transparent for regular users as noted by @infinisil in https://github.com/NixOS/org/pull/99#issuecomment-2932962410. Being able to look at the rule as a non-owner, allows to make much better suggestions for changes (for example for #116, #118, #119).
- Rulesets can be exported/imported as JSON. This allows much better collaboration on them, even without #40 for settings / rulesets. I could recreate the rule in my fork, make the changes I'm proposing, and export them as JSON. Those could even be reviewed as a diff and then imported.
- Rulesets are much more flexible, especially in terms of which branches to target (supporting exclusions, too). This is required for #118 for example.

We currently have 56 non-archived repos in the org. I briefly went through a random subset: A lot of repos don't have any branch protection (basic branch protection to prevent force pushes for the main branch should probably be added in most of them?) and I assume that many others only have very basic rules. My suggestion on how to approach this:
- Define a very simple, sensible default, for example (target: default branch, restrict deletion, block force pushes) and import this to all repos without protection.
- Import the same to those repos with matching classic rules right now.
- Deal with the rest manually.

I'm especially interested in the branch protection rules for `nixpkgs`, of course. As a starter, I could prepare a JSON export of the current configuration... **iff** an org owner / nixpkgs admin could provide screenshots of the current classic configuration...

Once we got this, I could then make much more specific suggestions for #116 and #118 (#119 is already very specific).

infinisil:
Sounds good to me!

Here's screenshots of all current Nixpkgs branch protections: https://gist.github.com/infinisil/33fdbb637e719a9ebeb46010b4761861

wolfgangwalther:
One low-hanging fruit: The branch protection rule "main" doesn't affect any branches and can be deleted. Edit: Although.. maybe the intent is to prevent *creation* of this branch, in which case it makes sense.

wolfgangwalther:
To make review easier, I opened a PR for the to-be-imported json files: #124.

wolfgangwalther:
[...]

PR #124, opened 2025-06-11
Switch nixpkgs to branch protection rulesets

wolfgangwalther:
This PR is not necessarily intended to be merged\*. It targets #120, but discussing code in a Pull Request is easier than in an issue. Those files are supposed to be *imported* into the nixpkgs branch protection rulesets section.

Those files have been created in my fork and were then exported. The only modification I made manually is in the `bypass_actors` section. Since I can't have any teams in my personal fork, I added those manually. Hopefully that works, it should according to API docs etc.

The Team ID's can be verified with:
```bash
gh api /orgs/NixOS/teams/channel-updaters

gh api /orgs/NixOS/teams/nixos-release-managers
```

The general structure is different compared to the classic branch protection rules, because the features are slightly different. The new rulesets can only define bypassers for the whole ruleset, not for some of the rules in it. Thus, the different rules must be separated to replicate the status-quo.

AFAICT, this should replicate the [status-quo](https://gist.github.com/infinisil/33fdbb637e719a9ebeb46010b4761861) exactly. Of course, careful review is required.

@infinisil 

---

\*If, however, we want to document those rules here, then they should be re-exported after import.

### DIFF
--- a/rulesets/export.bash
+++ b/rulesets/export.bash
@@ -1,0 +1,10 @@
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash -p gh jq
+set -euo pipefail
+
+REPO="${1:?Please pass the repository name as the only argument.}"
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+gh api "/repos/NixOS/$REPO/rulesets" --jq '.[] | "\(._links.self.href) \(.name)"' \
+	| xargs -n2 sh -c "gh api \$0 | jq 'del(.node_id, ._links)' > $REPO/\$1.json"
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,6 +11,7 @@
 /npins @NixOS/org
 /default.nix @NixOS/org
 /shell.nix @NixOS/org
+/rulesets @NixOS/org
 
 /doc/org-repo.md @NixOS/steering
 /doc/discourse.md @NixOS/steering @mweinelt @picnoir @lassulus @jtojnar @ctheune @dpausp
--- a/rulesets/README.md
+++ b/rulesets/README.md
@@ -1,0 +1,9 @@
+This directory contains JSON exports of [branch protection rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets) for repositories in the NixOS org.
+
+They are not managed automatically, but can easily be imported by org owners.
+To propose changes to branch protection rules, you can open a Pull Request.
+
+After making changes to those rulesets, please export them again.
+This can be done with the `./export.bash` script.
+For example, to export the rulesets for the `nixpkgs` repo, call `rulesets/export.bash nixpkgs`.
+While the rulesets endpoint can be read by non-owners, the export must be done by an org-owner to include the `bypass_actors` field.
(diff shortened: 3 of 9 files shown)


infinisil (approved):
Thank you! I've reviewed this and can confirm that this should match the status quo!

Since <del>I'm pretty sure</del> [we can](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets#about-rulesets-and-protected-branches) run both rulesets and branch protections simultaneously, I'll go ahead and import these right now (while doing another sanity check before saving). I'll update the PR with the exported ones then.

After that, let's do a brief check before getting another approval from an org admin to turn off branch protections.

infinisil:
@wolfgangwalther The only notable difference after importing is [this](https://github.com/NixOS/org/pull/124/commits/334e21adebd7627129b7c96a837291b66c6f332d#diff-be3dbfc39193dd5325eb0da8795417660028dfc018a316526ccc55721ff1321dL19-R19). Do you know what that's about?

wolfgangwalther:
> The only notable difference after importing is [this](https://github.com/NixOS/org/pull/124/commits/334e21adebd7627129b7c96a837291b66c6f332d#diff-be3dbfc39193dd5325eb0da8795417660028dfc018a316526ccc55721ff1321dL19-R19). Do you know what that's about?

Hrhr... `false` is the default here and I don't think I changed it. I *do* remember, that I clicked on the "Hide additional settings" button in the "restrict updates" section. Maybe that caused it to be saved differently for me somehow, aka saving the default explicitly?

Edit: To clarify, because the name is misleading, it's this setting:

> Allow fork syncing
>    Branch can pull changes from its upstream repository

(you can see the id via devtools)

infinisil:
Ah I see, there's no additional settings for organisations!

I invited you to my https://github.com/orgs/Infinisil-s-Test-Organization in case you want to mess around with org settings and don't want to create your own test org.

wolfgangwalther:
Of course, that makes sense. Maybe not because it's in an organization, but because the original *is not a fork*, so the setting doesn't make sense.

wolfgangwalther:
I added a simple README to the `rulesets/` folder and an `export.bash` script that makes it simpler to do the export. The export via the API will contain a few new fields, could you run it once, @infinisil? I can't do it, because the `bypass_actors` field is not available for non-owners.

The latter is also an argument why to keep the exports checked into the repo. The rules might be visible for everyone through the branches tab in the repos, but this does not contain the bypassing actors.

rulesets/export.bash
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash -p gh jq
+set -euo pipefail
+
+REPO="${1:?Please pass the repository name as the only argument.}"
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+gh api /repos/NixOS/nixpkgs/rulesets --jq '.[] | "\(._links.self.href) \(.name)"' \
```
infinisil:
```suggestion
gh api /repos/NixOS/"$REPO"/rulesets --jq '.[] | "\(._links.self.href) \(.name)"' \
```

wolfgangwalther: Ah, thanks, copy&pasto from a later test..

zimbatm (approved):
nice, thanks for helping make the project more transparent

wolfgangwalther:
> getting another approval from an org admin to turn off branch protections.

I think we can do that now.

Once this is merged, I can create PRs to propose actual *changes* to the branch protection rules.

infinisil:
Deleted the branch protections now, so we only rely on rulesets now!

PR #129, opened 2025-06-18
github: add surveys repository

djacu:
Add a new surveys repository for centralizing questions used in community and other official surveys, the data acquired from the survey and the code to process survey data into meaningful visuals.

### DIFF
--- a/doc/github.md
+++ b/doc/github.md
@@ -68,6 +68,12 @@ The home of the branding guide and media kit.
 
 The [marketing team](https://nixos.org/community/teams/marketing/) controls this repository.
 
+### [surveys](https://github.com/NixOS/surveys)
+
+The home of survey questions, historical data, and processing code.
+
+The [marketing team](https://nixos.org/community/teams/marketing/) controls this repository.
+
 ## Groups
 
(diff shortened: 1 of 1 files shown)


pbsds:
Is the data to be made public anonymized?

djacu:
@pbsds 
If I recall correctly, the data we collected from the NixCon NA 2024 survey was already anonymized. I have not see the raw Nix Community survey data from the past 2 years, but if it is not anonymized directly from lime survey, then it would be made so before publishing.

@fricklerhandwerk @GuillaumeDesforges since you two worked on the surveys in the past, do you see any issue with publishing the answer data if it is anonymized?

fricklerhandwerk:
I would absolutely not publish the raw data, it's too fine-grained to be meaningfully anonymous. We didn't do it in the past either. But having a normalized public representation of the aggregate results would be super valuable to analyze trends.

djacu:
Ok no raw data then. Only normalized data if possible.

pbsds (approved):

infinisil (approved):
Done, was approved by @Lassulus on Matrix

PR #131, opened 2025-06-19
rulesets/nixpkgs: remove exception for force push for release team

wolfgangwalther:
We currently have two exceptions for the @NixOS/nixos-release-managers team in the Nixpkgs branch protection rules:
- They can force push to staging branches.
- ~~They can push to release branches without Pull Requests.~~ (valid use-case)

AFAICT, neither is required for the release process documented at https://nixos.github.io/release-wiki/Branch-Off.html. Those exceptions are most likely caused by the old branch protection rules, which were not as flexible as the new rulesets.

Note: Branch *creation* does not require "pushing to release branches". This only applies after the branch is created and more commits should be pushed to it. I just confirmed that in my fork to be sure.


leona-ya:
This is not correct with how we did the last release: 

1. At branch-off, the beta commit is currently pushed directly. Here I don't have much problem with changing that tho
2. https://nixos.github.io/release-wiki/Final-Release.html#at-final-release-time: The final release commit is currently pushed to the protected branch. We _might_ be able to do that with a PR, but for me it seems cursed to tag a merge commit as "the release commit" or a commit that was merged with a merge commit as this might has a non-consistent state.

I'm happy to discuss whether my option here is strange, but I think we should do that and not remove rights before that.

wolfgangwalther:
Ah, I didn't read through the Final Release part. That's the reason for that exception, no problem with that. I'll happily drop this commit and leave it at the removal of the force push exception.

wolfgangwalther:
I documented the reason for the exception in `rulesets/nixpkgs/README.md` instead.

infinisil (approved):
Looking good to me! I'll go ahead with the implementation of this

Issue #118, opened 2025-06-01
nixpkgs branch protection rules: prevent creation of new branches!


wolfgangwalther:
More often than not, branches are created in the nixpkgs repo by accident, when they should actually be created in committers forks.

This leads multiple problems:
- The fork's "delete after merge" policy doesn't kick in.
- The nixpkgs repo's branches are cluttered wild old, abandoned stuff.
- Sometimes people accidentally create names matching an existing branch protection rule. Those branches **can't be cleaned up** (by ordinary committers like me, at least). Examples: `staging.patchShebangs`, `release-18.09-firefox64`, `staging-patchelf`, .. there are more.

Imho, there are exactly 3 reasons to have branches in the nixpkgs upstream repo itself:
- channel branches starting with `nixpkgs-` and `nixos-`.
- backport branches, created by CI starting with `backport-`.
- development branches: `master`, `release-xx.yy`, `staging` (...), `haskell-updates`, `python-updates`...

There is a bit of inconsistency in the development branches right now. For example, we also have `r-updates`, which is a protected branch, but is not listed in `ci/request-reviews/dev-branches.txt` (and the relevant workflows). And we have `perl-updates`, which is **not** a protected branch, but seems to be used in the same way.

To prevent us from introducing more of that, I would like to suggest the following:
- Create a new branch protection rule, which matches **all** branches and does **not** allow branch **creation**. The backport app and the release team would be added as those who can bypass this rule.
- Somebody with owner permission to go through the branch list and look for those **accidentally protected branches** (mentioned above, there are more - they all start with a prefix of release/staging, two also with nixos-test-) and delete them.
- Clarify the status of `r-updates` and `perl-updates` - and possibly other "development branches" that might be needed to collaborate together on a topic. Those should all follow the same CI principles and be protected similarly. For example that means they can be cherry-picked for backports from etc.

We'd then be fading out the still open PRs over time, by deleting the relevant branches when the PR has been closed/merged.

mweinelt:
We regularly collaborate on the home-assistant branch, as we need to merge open PRs into it. Once merged we tend to delete it until we need it again.

wolfgangwalther:
> We regularly collaborate on the home-assistant branch, as we need to merge open PRs into it.

In that case, it should be an official "development branch", so have a protection rule and be added to the respective workflows.

(If we want to go full consistency, maybe name it `home-assistant-updates`, too?)

> Once merged we tend to delete it until we need it again.

[...]

PR #132, opened 2025-06-19
rulesets/nixpkgs: disallow creation of new branches

wolfgangwalther:
This disallows the creation of new branches in the nixpkgs repo with the following exceptions:
- `backport-`, `revert-` and `wip-` branches can be created by anyone.
- `nixos-` and `nixpkgs-` branches can be created by the channel updaters.
- `release-` and `staging-` branches can be created by the release team.

This does not affect any existing branches, which we can fade out over time instead.

This is work towards #118, although I wouldn't close that issue with this, yet. There's more ideas in there, some of which I'd need to move to the nixpkgs repo's CI as well. But also the clarification of "development branches" for some other branch protection rules.

### DIFF
--- a/rulesets/nixpkgs/no-creation.json
+++ b/rulesets/nixpkgs/no-creation.json
@@ -7,9 +7,17 @@
   "enforcement": "active",
   "conditions": {
     "ref_name": {
-      "exclude": [],
+      "exclude": [
+        "refs/heads/backport-*",
+        "refs/heads/nixos-*",
+        "refs/heads/nixpkgs-*",
+        "refs/heads/release-*",
+        "refs/heads/revert-**",
+        "refs/heads/staging-*",
+        "refs/heads/wip-*",
+      ],
       "include": [
-        "refs/heads/main"
+        "refs/heads/**"
       ]
     }
   },
--- a/rulesets/nixpkgs/no-creation.json
+++ b/rulesets/nixpkgs/no-creation.json
@@ -14,10 +14,10 @@
         "refs/heads/release-*",
         "refs/heads/revert-**",
         "refs/heads/staging-*",
-        "refs/heads/wip-*",
+        "refs/heads/wip-*"
       ],
       "include": [
-        "refs/heads/**"
+        "~ALL"
       ]
     }
   },
--- a/rulesets/nixpkgs/no-creation-except-channel-updaters.json
+++ b/rulesets/nixpkgs/no-creation-except-channel-updaters.json
@@ -1,0 +1,29 @@
+{
+  "name": "no-creation-except-release-team",
+  "target": "branch",
+  "source_type": "Repository",
+  "source": "NixOS/nixpkgs",
+  "enforcement": "active",
+  "conditions": {
+    "ref_name": {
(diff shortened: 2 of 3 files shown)


wolfgangwalther:
The related branches have also been documented here: https://github.com/NixOS/nixpkgs/blob/master/ci/README.md#branch-classification. We marked "all other branches", which would be forbidden to be created by these branch protection rules here, as "deprecated".

zimbatm:
Did you think how new branches could still be created once this PR is merged?

There might be some scenarios where maintainers need to create semi-long-lived branches, so they can be built on Hydra. Let's say a new Gnome release for example. When that happens, the maintainers need to know what process they need to use to get that branch created.

wolfgangwalther:
The idea is to use the `wip-` namespace for that - and anyone could create branches there.

Where would you suggest would this best be documented? In `ci/README.md` in the nixpkgs repo or somewhere else?

infinisil (approved):
This sounds good to me, but can we can approval from another @NixOS/org owner?

rulesets/nixpkgs/no-creation.json
```diff
@@ -7,9 +7,17 @@
   "enforcement": "active",
   "conditions": {
     "ref_name": {
-      "exclude": [],
+      "exclude": [
+        "refs/heads/backport-*",
+        "refs/heads/nixos-*",
+        "refs/heads/nixpkgs-*",
+        "refs/heads/release-*",
+        "refs/heads/revert-**",
```
infinisil, line 15: Should this be

```suggestion
        "refs/heads/revert-*",    
```

Or is there no difference?

wolfgangwalther: There is a difference: `**` will include `/`. Aka when you have a branch called `revert-123456-feat/stuff`, `revert-*` will not match, but `revert-**` will.

Since the `revert-` branches are auto-created from the PR number *and the head branch of the PR*, which was provided by a regular contributor, this can use the `/` format.

I'm OK with not allowing these for `wip-` and `backport-` prefixes, but for the above reason we must support them for `revert-`.

Edit: This is a moment where I miss the ability to add comments to `.json`...

Lassulus (approved):

infinisil:
Applied!

rulesets/nixpkgs/no-creation.json
wolfgangwalther: > There is a difference: `**` will include `/`. Aka when you have a branch called `revert-123456-feat/stuff`, `revert-*` will not match, but `revert-**` will.

That's what the docs tell me, at least: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/creating-rulesets-for-a-repository#using-fnmatch-syntax

But it's apparently not that easy, because @philiptaron [just hit this](https://github.com/NixOS/nixpkgs/pull/420319#issuecomment-3079769010) when trying to revert a r-ryantm. The branch is called `auto-update/re2` - and is blocked by our new rule.

I reproduced this in my fork as well. Regular branches work well, but those with `/` don't.

Issue #130, opened 2025-06-19
nixpkgs branch protection rules: require status checks to pass


wolfgangwalther:
https://github.com/NixOS/nixpkgs/pull/417957 has the base to allow enabling required status checks efficiently, thus I propose to do that.

Requiring status checks would help prevent accidental merges. To be clear, they are **not** about lack of trust. Instead, they are about **protecting committers**. Nobody wants to break CI for others by accidentally merging something that actually fails CI. Also, those checks are a prerequisite for **merge queues**.

A little breakdown of how they work:
- The branch protection rule specifies the *name* of a status check. This will appear as "pending" in all PRs immediately after the rule is activated. It looks like this:

  ![Image](https://github.com/user-attachments/assets/c52e6045-7106-4d06-9188-bbd2a1401de6)

- Once a job with this name runs, it will replace that pending status check with either "success/skipped" or "failure". Skipped counts as success. When it's successful, a merge is allowed. When it's failed, the merge is blocked. In the UI, the merge button is disabled (similar to when a PR is merge conflicted). When merging from the command line, it shows like this:

  ```
   remote: error: GH013: Repository rule violations found for refs/heads/master.
   remote: Review all repository rules at https://github.com/wolfgangwalther/nixpkgs/rules?ref=refs%2Fheads%2Fmaster
   remote:
   remote: - Required status check "required to merge" is expected.
   remote:
   To github.com:wolfgangwalther/nixpkgs.git
    ! [remote rejected]           HEAD -> master (push declined due to repository rule violations)
   error: failed to push some refs to 'github.com:wolfgangwalther/nixpkgs.git'
   ```

- The branch protection rule can have a number of bypassers set up. The teams/roles with this privilege can merge a PR despite some required checks failing. In the web UI, it looks like this:
  
  ![Image](https://github.com/user-attachments/assets/adec584b-e8e3-4f68-9f57-665757589b4a)

- When merging from the command line with bypass privileges, the **push will always succeed**. If a branch protection rule failed, it looks like this:

  ```
  remote: Bypassed rule violations for refs/heads/master:
  remote:
  remote: - Required status check "required to merge" is expected.
  remote:
  To github.com:wolfgangwalther/nixpkgs.git
     f68c81b1b9b4..a6086158c33b  HEAD -> master
  ```

Now, the question becomes: Which bypassers do we want to allow?

[...]

PR #134, opened 2025-06-25
rulesets/nixpkgs: introduce required status checks

wolfgangwalther:
This enables the "Required Status Checks" feature for Nixpkgs' development branches.

At this stage, all nixpkgs committers can bypass the checks. See discussion in #130.

Notes:
- I intentionally left out `python-updates` and `r-updates`, which both have "no-delete" rules, but not "no-force-push". Status of those branches is discussed in #118.
- The teams ID can be confirmed with: `gh api /orgs/NixOS/teams/nixpkgs-committers`.
- #130 also mentions a "allow reverts" step, but this can be done *after* enabling the ruleset, since all committers can bypass anyway.

cc @MattSturgeon

### DIFF
--- a/rulesets/nixpkgs/required-status-checks.json
+++ b/rulesets/nixpkgs/required-status-checks.json
@@ -1,4 +1,5 @@
 {
+  "id": 6624327,
   "name": "required-status-checks",
   "target": "branch",
   "source_type": "Repository",
@@ -11,7 +12,7 @@
         "~DEFAULT_BRANCH",
         "refs/heads/release*",
         "refs/heads/staging*",
-        "refs/heads/haskell-updates",
+        "refs/heads/haskell-updates"
       ]
     }
   },
@@ -30,6 +31,8 @@
       }
     }
   ],
+  "created_at": "2025-07-10T22:48:00.187+02:00",
+  "updated_at": "2025-07-10T22:48:00.187+02:00",
   "bypass_actors": [
     {
       "actor_id": 203427,
@@ -37,5 +40,5 @@
       "bypass_mode": "always"
     }
   ],
-  "current_user_can_bypass": "never"
+  "current_user_can_bypass": "always"
 }
--- a/rulesets/nixpkgs/required-status-checks.json
+++ b/rulesets/nixpkgs/required-status-checks.json
@@ -1,0 +1,41 @@
+{
+  "name": "required-status-checks",
+  "target": "branch",
+  "source_type": "Repository",
+  "source": "NixOS/nixpkgs",
+  "enforcement": "active",
+  "conditions": {
+    "ref_name": {
+      "exclude": [],
+      "include": [
+        "~DEFAULT_BRANCH",
+        "refs/heads/release*",
+        "refs/heads/staging*",
+        "refs/heads/haskell-updates",
+      ]
+    }
+  },
+  "rules": [
+    {
+      "type": "required_status_checks",
+      "parameters": {
+        "strict_required_status_checks_policy": false,
+        "do_not_enforce_on_create": true,
+        "required_status_checks": [
+          {
+            "context": "no PR failures",
+            "integration_id": 15368
+          }
+        ]
+      }
+    }
+  ],
+  "bypass_actors": [
+    {
+      "actor_id": 203427,
+      "actor_type": "Team",
+      "bypass_mode": "always"
+    }
+  ],
+  "current_user_can_bypass": "never"
+}


rulesets/nixpkgs/required-status-checks.json
```diff
@@ -0,0 +1,41 @@
...
+  "source_type": "Repository",
+  "source": "NixOS/nixpkgs",
+  "enforcement": "active",
+  "conditions": {
+    "ref_name": {
+      "exclude": [],
+      "include": [
+        "~DEFAULT_BRANCH",
+        "refs/heads/release*",
+        "refs/heads/staging*",
```
MattSturgeon, line 14: Note: if/when merge queues are enabled, they do not support wildcard branch patterns (for some reason)

MattSturgeon (approved):
- Diff LGTM
- Committers team id verified via rest api

> This directory contains JSON exports of [branch protection rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets) for repositories in the NixOS org.
> 
> They are not managed automatically, but can easily be imported by org owners.
> To propose changes to branch protection rules, you can open a Pull Request.

For a second, I thought there was some magic github org-repo that'd could declaratively manage repo settings...

infinisil (approved):
While I can't actually import this ruleset directly (gives an error), I can manually create it and then re-export it. Can I get another approval from an @NixOS/org owner for this? :)

@wolfgangwalther Thanks a lot for all this work, very appreciated!

wolfgangwalther:
> While I can't actually import this ruleset directly (gives an error)

Interesting, what's wrong with it, I wonder? I guess we will see after export ;)

Lassulus (approved):

infinisil:
Applied!

infinisil:
Btw the problem with the importing was just that the JSON list had an extra `,` at the end :laughing:

mweinelt:
What prevents us from enabling merge queues now?

infinisil:
We could probably try that, I have the option to enable this for the ruleset:

Screenshot 2025-07-10 at 23-00-33 Settings · Rulesets · nixpkgs

infinisil:
These are also options, repo-wide:

image

mweinelt:
Yes, auto-merge would make reviewing and merging async. That would be super helpful. And you can cancel until it is merged.

Actually, I care more for auto-merge than merge-queues. Merge queues would be an experiment, auto-merges should be fairly safe.

wolfgangwalther:
re merge queues: before we can enable these, we need to split the ruleset etc. - we can't require them for the staging branches right now, because all our periodic merges need to bypass that etc.

re auto-merge: that's surely something we can try. we just need to consider that it won't wait for ofborg - it will only wait for eval and the other basic checks.

MattSturgeon:
> These are also options, repo-wide:
> 
> image

FYI "Auto Merge" can be enabled either per-repo or per-ruleset. Also, I believe requiring "Merge Queues" implicitly enables "Auto Merge" for that ruleset, but I could be wrong.

> What prevents us from enabling merge queues now?

We also need to add the `on: merge_group` event to the PR workflow; or add a separate workflow that runs `on: merge_group` but _also_ has a `no PR failures` job.

Although I'm also looking forward to Merge Queues, I'd personally like to see the Required Status Checks working well in production before we rush into requiring a queue.

EDIT: We'd also need to consider how the ruleset targets branches, because Merge Queues cannot be used on rulesets that target wildcard branch patterns. E.g. we could have a "base" ruleset that uses wildcards, alongside a separate "overlay" ruleset that requires merge queues on explicitly named branches.

mweinelt:
This regressed our periodic merge CI.

https://github.com/NixOS/nixpkgs/actions/runs/16222366363/job/45805853015

emilazy:
`staging*` branches should probably be omitted from these kinds of changes by default since the workflow is significantly different. (It would be nice to make the workflow less different, but the reality is that direct pushes to `staging-next` etc. are still common.)

infinisil:
Staging is now not included anymore: https://github.com/NixOS/org/pull/139

wolfgangwalther:
As mentioned in #139, I think we should keep this on staging, but allow the nixpkgs-ci app to bypass this ruleset as well. We actually discussed this issue already in https://github.com/NixOS/org/issues/130#issuecomment-3004390494 - but I entirely missed that we need nixpkgs-ci to be a bypasser as well.

PR #137, opened 2025-07-11
rulesets/nixpkgs: delete accidentally created protected branches

wolfgangwalther:
These branches had been created with a protected prefix in the past - and thus accidentally became protected. This also meant that no pushes were possible eventually - and thus these branches were quickly abandoned.

This PR intends to add these temporarily to the ruleset as exceptions - to then delete the branches immediately. Thus, this PR doesn't need to actually be merged. Importing the ruleset, deleting the branches, and then reverting the ruleset to its previous state should be enough.

Now that #132 prevents the accidental creation of new cases, we can do this.

### DIFF
--- a/rulesets/nixpkgs/no-delete.json
+++ b/rulesets/nixpkgs/no-delete.json
@@ -7,7 +7,14 @@
   "enforcement": "active",
   "conditions": {
     "ref_name": {
-      "exclude": [],
+      "exclude": [
+        "refs/heads/nixos-test-staging",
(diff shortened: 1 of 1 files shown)


infinisil:
Done as suggested, the branches are now deleted, thank you!

wolfgangwalther:
@infinisil I found one more, which followed an odd naming scheme that didn't match my search: [staging.patchShebangs](https://github.com/NixOS/nixpkgs/tree/staging.patchShebangs)

infinisil:
Done :)

PR #142, opened 2025-07-16
rulesets/nixpkgs: allow reverting r-ryantm branches via UI

wolfgangwalther:
These branches contain a `/` as in `auto-update/re2`. While `**` is supposed to match `/`, the glob syntax is still not easy to get right.

Tested this in my fork, this works for both branches with `/` and without `/`.

Reported by @philiptaron in https://github.com/NixOS/nixpkgs/pull/420319#issuecomment-3079769010.

@infinisil can you fix this?


philiptaron (approved):
Thanks for your attention here @wolfgangwalther. Appreciate you.

infinisil (approved):
Thanks, applied!

PR #144, opened 2025-07-22
rulesets/nixpkgs: allow dependabot branch creation

wolfgangwalther:
Dependabot can't create branches anymore to update github-actions: https://github.com/NixOS/nixpkgs/network/updates

Broke in #132.

I hope this fixes it, but the double star glob was confusing the last time, so let's see...

PR #146, opened 2025-07-24
rulesets/nixpkgs: fix branch creation rules

wolfgangwalther:
Allowing any number of `/` needs the `**/*` syntax.

While at it with the default ruleset, I also fixed https://github.com/NixOS/org/pull/144#discussion_r2222216286. This time, I created these rules in my fork and ran various `gh api /repos/<org>/nixpkgs/rules/branches/....` request against it. This now works with any number of `/`.

### DIFF
--- a/rulesets/nixpkgs/no-creation.json
+++ b/rulesets/nixpkgs/no-creation.json
@@ -15,8 +15,9 @@
         "refs/heads/staging-*",
         "refs/heads/wip-*",
         "refs/heads/revert-*",
-        "refs/heads/revert-*/**",
-        "refs/heads/dependabot/**"
+        "refs/heads/wip-*/**/*",
+        "refs/heads/revert-*/**/*",
+        "refs/heads/dependabot/**/*"
       ],
       "include": [
         "~ALL"
@@ -29,7 +30,7 @@
     }
   ],
   "created_at": "2025-06-11T20:44:19.501+02:00",
-  "updated_at": "2025-07-22T13:05:30.263+02:00",
+  "updated_at": "2025-08-11T16:26:43.654+02:00",
   "bypass_actors": [],
   "current_user_can_bypass": "never"
 }


rulesets/nixpkgs/no-creation.json
```diff
@@ -15,8 +15,8 @@
         "refs/heads/staging-*",
         "refs/heads/wip-*",
```
philiptaron, line 16: Should we convert over `wip-*` too?

wolfgangwalther: I'd not change these. This only allows people to create `wip-...` branches without slashes, but that should be fine. The only reason we need this for the revert and dependabot branches is, because we can't control their names.

philiptaron: I like the idea of `wip-issue-12345/description-of-approach` style of branches; that's what I use. It seems an arbitrary restriction which is why I would prefer `wip-*/**/*` for consistency sake.

wolfgangwalther: Added `wip-*/**/*`. Although I still think your example is way too specific for a collaborative branch, that gets merged into with PRs.

I expect names like `wip-python-updates`, `wip-perl-packages`, `wip-modular-services`, stuff like that - and then PRs targeting these.

For specific issues, your fork is the right place.

MattSturgeon: > For specific issues, your fork is the right place.

Generally I agree, although there are exceptions to every rule. Hypothetically, I could imagine a particularly challenging issue that required some collaboration. But this seems very niche and unlikely.

rulesets/nixpkgs/no-creation.json
```diff
@@ -14,9 +14,10 @@
         "refs/heads/release-*",
         "refs/heads/staging-*",
         "refs/heads/wip-*",
+        "refs/heads/wip-*/**/*",
         "refs/heads/revert-*",
-        "refs/heads/revert-*/**",
-        "refs/heads/dependabot/**"
+        "refs/heads/revert-*/**/*",
```
MattSturgeon, line 19: Q: why do we need `revert-*` _and_ `revert-*/**/*` ?

Wouldn't `revert-**` be sufficient?

(Same applies to all lines)

wolfgangwalther: The `/**/*` syntax covers 1+ `/`, but not 0, AFAICT. Thus we need to cover the case without `/` explicitly.

MattSturgeon: > The `/**/*` syntax covers 1+ `/`, but not 0

Yes, but the `**` pattern covers 0 or more, so can we simplify to `foo-**` (not `foo-*/**` or `foo-*/**/*`). Maybe I'm missing something?

Also, some of these lines have `foo-*/**/*` which I read as **2** or more `/`, not **1** or more.

wolfgangwalther: > Yes, but the `**` pattern covers 0 or more, so can we simplify to `foo-**` (not `foo-*/**` or `foo-*/**/*`). Maybe I'm missing something?

I tried this and this didn't work. That's precisely why I came up with this PR. See https://github.com/NixOS/org/pull/132#discussion_r2191055845 and #144.

If you want to try yourself, take a look at https://github.com/NixOS/org/pull/144#issuecomment-3102286689 on how to test this in practice. If you can come up with something simpler that works for all cases, cool.

> Also, some of these lines have `foo-*/**/*` which I read as **2** or more `/`, not **1** or more.

No, `**/*` matches 0 or more. Thus, `/**/*` matches 1 or more. I'm not exactly sure whether `revert-**/*` would work. I think that's the only combination that I didn't explicitly try, I think.

wolfgangwalther: > I'm not exactly sure whether `revert-**/*` would work. I think that's the only combination that I didn't explicitly try, I think.

Just tried that as well and that doesn't work. The current proposal in this PR is the only one I can come up with that works for all cases.

MattSturgeon: Apparently GitHub is using Ruby's [fnmatch](https://ruby-doc.org/core-2.5.1/File.html#method-c-fnmatch). Some of the examples are strange:

```ruby
File.fnmatch('**.rb', 'main.rb')     #=> true
File.fnmatch('**.rb', './main.rb')   #=> false
File.fnmatch('**.rb', 'lib/song.rb') #=> true
```

> If you want to try yourself, take a look at [#144 (comment)](https://github.com/NixOS/org/pull/144#issuecomment-3102286689) on how to test this in practice.

Thanks. Tested myself too, and I confirm your findings. I would _expect_ it to be possible, but I can't find a working pattern either.

In my testing:

1. `foo-**` does not match any `/`, only matching e.g. `foo-bar` not `foo-b/ar`
2. `foo-**/**` only matches a single `/`, matching `foo-b/ar` not `foo-bar` or `foo-b/a/r`
3. `foo-**/*` also only matches a single `/`

It seems like `fnmatch` is only treating `**` as a recursive glob when it is on its own (e.g. between `/`), when `**` is part of a wider pattern, it seems to be parsed as two separate `*` wildcards. It also seems strange that `/**` only matches a single `/` and isn't recursive...

I'm sure POSIX globbing doesn't behave like this?

---

Regardless, I think this resolves my initial question...

infinisil (approved):
Applied!

wolfgangwalther:
Thanks! It works - we are now getting Dependabot updates again.

PR #149, opened 2025-08-05
rulesets/nixpkgs: add merge queue for lints on master

wolfgangwalther:
This requires https://github.com/NixOS/nixpkgs/pull/431146 and enables a very basic merge queue for the master branch in nixpkgs. At this stage, this only runs the "lint" group of checks: treefmt, nixpkgs-vet and the parse check. We should do this right now, because the recent nixfmt 1.0.0 update can easily cause nixfmt regressions after merging a seemingly OK PR - where nixfmt ran before the update, but the would need to run again afterwards.

We're not including all of the other checks, yet, because any change included here means it can't be bypassed anymore. We'll need to investigate whether we might still need to be able to do this later on, for now I added the CI team as by-passers as a safety net

This includes the commit from #146, to avoid merge conflicts. Also, we should probably wait on that to be merged anyway, because it currently causes dependabot to not do updates. This means we're not running Nix 2.30 in GHA, yet, which means we are still subject to random nixpkgs-vet failures. These go away after rerunning the job, but would still kick the PR out of the queue at first. So we might just as well wait for this update to be in.

cc @NixOS/nixpkgs-ci

### DIFF
--- a/rulesets/nixpkgs/require-merge-queue.json
+++ b/rulesets/nixpkgs/require-merge-queue.json
@@ -1,0 +1,36 @@
+{
+  "name": "require-merge-queue",
+  "target": "branch",
+  "source_type": "Repository",
+  "source": "NixOS/nixpkgs",
+  "enforcement": "active",
+  "conditions": {
+    "ref_name": {
+      "exclude": [],
+      "include": [
+        "~DEFAULT_BRANCH"
+      ]
+    }
+  },
+  "rules": [
+    {
+      "type": "merge_queue",
+      "parameters": {
+        "merge_method": "MERGE",
+        "max_entries_to_build": 5,
+        "min_entries_to_merge": 1,
+        "max_entries_to_merge": 5,
+        "min_entries_to_merge_wait_minutes": 5,
+        "grouping_strategy": "ALLGREEN",
+        "check_response_timeout_minutes": 60
+      }
+    }
+  ],
+  "bypass_actors": [
+    {
+      "actor_id": 203427,
+      "actor_type": "Team",
+      "bypass_mode": "pull_request"
+    }
+  ]
+}
--- a/rulesets/nixpkgs/no-creation.json
+++ b/rulesets/nixpkgs/no-creation.json
@@ -17,7 +17,8 @@
         "refs/heads/revert-*",
         "refs/heads/wip-*/**/*",
         "refs/heads/revert-*/**/*",
-        "refs/heads/dependabot/**/*"
+        "refs/heads/dependabot/**/*",
+        "refs/heads/gh-readonly-queue/**/*"
       ],
       "include": [
         "~ALL"


wolfgangwalther:
The "check references" job fails, but this seems unrelated to this PR.

rulesets/nixpkgs/require-merge-queue.json
```diff
@@ -0,0 +1,36 @@
...
+  "name": "require-merge-queue",
+  "target": "branch",
+  "source_type": "Repository",
+  "source": "NixOS/nixpkgs",
+  "enforcement": "active",
+  "conditions": {
+    "ref_name": {
+      "exclude": [],
+      "include": [
+        "~DEFAULT_BRANCH"
```
MattSturgeon, line 11: Just for the record: we will _never\*_ be able to add wildcard branches to the include/exclude lists for this ruleset, as GitHub currently doesn't support requiring Merge Queues on a ruleset with wildcard patterns.

All listed branches must either be the default branch, or a non-wildcard pattern.

This means we can't add `refs/heads/release-*` and would instead have to add (e.g.) `refs/heads/release-25.05`.

_\*At least, until GitHub adds support._

rulesets/nixpkgs/require-merge-queue.json
```diff
@@ -0,0 +1,36 @@
...
+      }
+    }
+  ],
+  "bypass_actors": [
+    {
+      "actor_id": 13362067,
+      "actor_type": "Team",
+      "bypass_mode": "always"
+    }
+  ]
```
MattSturgeon: > We're not including all of the other checks, yet, because any change included here means it can't be bypassed anymore. We'll need to investigate whether we might still need to be able to do this later on, for now I added the CI team as by-passers as a safety net

This seems reasonable for now. Anyone bypassing the Merge Queue will a) reset the queue, affecting other PRs and b) almost certainly break CI on master.

---

Have you validated how this interacts with the other ruleset's bypass? E.g. can a committer bypass `pr.yml` and enqueue a PR?

I assumed that bypassing any rule meant all rules were ignored, even if they come from separate rulesets. I.e. I assume to merge a failing PR, you must also bypass the queue and you cannot enqueue a failing PR? 🤔 

So I theorize this may accidentally prevent normal committers from bypassing _anything_ being merged into master.

MattSturgeon (commented):
Excited to get merge queues soon 👀

rulesets/nixpkgs/require-merge-queue.json
wolfgangwalther: Yeah, that might not be too bad, though. In the current release process, [the release team needs to push directly to the release branch anyway](https://github.com/NixOS/org/pull/131#issuecomment-2988596884), so the merge queue would require them to do that differently. There could be a step in the release process to have the merge queue enabled for the new branch, explicitly.

rulesets/nixpkgs/require-merge-queue.json
wolfgangwalther: > Have you validated how this interacts with the other ruleset's bypass? E.g. can a committer bypass `pr.yml` and enqueue a PR?

I have not explicitly tested this, no, but ...

> I assumed that bypassing any rule meant all rules were ignored, even if they come from separate rulesets 🤔

... my experience with separate rulesets so far has been, that the bypassers only explicitly bypass the rules they are configured for. This is the big advantage compared to classic branch protection rules. I tested it with other rules a while ago, ...
 
> So I theorize this may accidentally prevent normal committers from bypassing _anything_ being merged into master.

... thus I don't expect this to happen. Committers should be able to bypass the required checks to hit the button, thus adding the PR to the queue. They won't be able to bypass the merge queue checks, so they will effectively never be able to bypass the lints. On master, that is.

MattSturgeon: Hm. I'm trying to test this in a test org+repo: https://github.com/MattSturgeonOrg/TestRepo/pull/1 (an org is needed to enable Merge Queues)

I have a ruleset that requires a status check (that'll never run), but can be bypassed by users with `write`: [Bypassable.json](https://github.com/user-attachments/files/21600866/Bypassable.json)

I have a second ruleset that requires merge queue: [Queue.json](https://github.com/user-attachments/files/21600868/Queue.json)

I have a "merge when ready" button, but no "bypass" tickbox:

Image

"Merge when ready" simply enables Auto Merge on the PR: 

Image

I also don't get any "bypass" checkbox if I manually fail the required status check:

Image

```sh
gh api /repos/MattSturgeonOrg/TestRepo/statuses/155a82d8f0c68c4cd5832899b5685916e7fbf2e2 \
  -f 'state=failure' \
  -f 'context=foo' \
  -f 'description=sent from gh api CLI'
```

wolfgangwalther: Ah, OK, I see what you mean now. I assume bypassing is still possible via CLI / API merge or so - but it seems like the UI just doesn't handle this case at all.

The UI effectively turns this into a "can't be bypassed".

What if you're allowed to bypass the queue? Does another option appear for you somehow?

MattSturgeon: > What if you're allowed to bypass the queue? Does another option appear for you somehow?

Yeah, that allows merging directly into the base branch:

image

wolfgangwalther: So essentially... GitHub assumes the "required status checks" to be the *same* for PR and Merge Queue. With that assumption, the "bypass the PR checks, but still run the queue" doesn't make sense.

We use the same name, but have different checks right now. Which is not really well supported, at least for bypassing.

Thus, we'll need to decide whether we're ready to take "everyone can bypass required status checks when merging into master" away from committers. Since nobody has reported in #130, so far, I'd say we're on the right track to do that. WDYT?

Edit: Ofc, we could also add the committer team as bypassers to the merge queue as well. Not sure whether we should, though.

MattSturgeon: Do we have any stats on how often anyone has bypassed (intentionally or otherwise)?

A manual `git push upstream HEAD:master` would bypass the ruleset(s), including bypassing the queue. I assume most of the time such a bypass would be an _unintended_ side-effect of preferring the CLI to the GUI.

I assume such a bypass will reset the merge queue and all queue groups. If this doesn't happen often, it's probably a non-issue.

Right now my instinct is that we just need to trial it for real; add all committers as bypassers, then wait to see how often legitimate and illegitimate bypasses happen in practice.

wolfgangwalther: > Do we have any stats on how often anyone has bypassed (intentionally or otherwise)?

I think it should be possible to get some stats via this endpoint: https://docs.github.com/en/rest/repos/rule-suites?apiVersion=2022-11-28#list-repository-rule-suites

I think this is only available to org/repo owners, though. I can't see anything there for upstream nixpkgs.

wolfgangwalther: > A manual `git push upstream HEAD:master` would bypass the ruleset(s), including bypassing the queue. I assume most of the time such a bypass would be an _unintended_ side-effect of preferring the CLI to the GUI.

I think this is the most likely thing to happen, yes. And I think that's why we should avoid adding everyone as bypassers.

MattSturgeon: We need to weigh up how often it will be necessary to bypass, how likely accidental bypasses are, and whether the CI team should be hands-on with such cases.

With this PR as-is, the CI team would have access to the "bypass" checkbox and be able to merge directly, skipping status checks _and_ the queue.

If bypasses are rare enough, then perhaps it is ok for committers to ping the CI team whenever a bypass is needed. I suppose we'd want to know about such issues _anyway_, so perhaps it does make sense to require us be pinned?

We could communicate this through the pinned issue and/or CONTRIBUTING documents.

emilazy: I think the CI team is too small a group. In the worst case we could end up with merges being impossible until someone comes to fix it, and CI has been a bit flaky lately. Probably the security team should be able to do it as well. Maybe we should start with the committers team and then shrink it.

(My comments on the Nixpkgs PR are relevant to the discussion of checks in the queue vs. on the PR.)

MattSturgeon: > I think the CI team is too small a group. In the worst case we could end up with merges being impossible until someone comes to fix it, and CI has been a bit flaky lately. Probably the security team should be able to do it as well. Maybe we should start with the committers team and then shrink it.

I agree adding the security team makes sense, assuming they're happy with that. I vaguely recalled a prior discussion where they hinted that they may not want the extra workload of being pinged to bypass CI issues?

Now that status checks have been required for a while, we can consider how often they've been bypassed. Hopefully those numbers will help us weigh up how many people should have bypass here.

To be clear, I'm fine with having all committers bypass.

But I see why @wolfgangwalther is keen to reduce the numbers; GitHub makes it too easy to accidentally bypass rulesets via the `git` CLI, which could be potentially inconvenient if that resets a large queue.

If we can't reach a consensus or get the relevant stats in time, then we can maintain the status quo (all committers can bypass) and re-review this later in another PR.

wolfgangwalther: While I agree that basically every committer should be able to bypass it, especially in the beginning, we are essentially disallowing the "merge via CLI" approach that some committers use by introducing a merge queue. And if we only soft-disable it, aka by still allowing it (because every push is a bypass), but having a queue reset every time, then that's quite counterproductive - especially when those merging via CLI don't notice it. Blocking them from merging would clearly show the problem with this approach, though.

I'd like to hear @vcunat's opinion on this, too, who earlier said is merging via CLI regularly.

vcunat: Basically all my nixpkgs merges are via CLI because that way I can *sign* them.

wolfgangwalther: For my understanding: By CLI you mean `git push` and not `gh pr merge`, correct?

Unfortunately, `git` merges are fundamentally not supported with merge queues. So there's two things where your feedback is needed, @vcunat:
- Should we implement merge queues at all? I guess those who discussed here so far all agree on that, but do you think they are not worth it, when we need to get rid of CLI merges at the same time?
- If we implement merge queues, would it be enough to just say "don't use CLI merges anymore" or should we rather block them?

vcunat: Yes, `git push`.

I haven't thought through the tradeoffs fully.  With mechanisms like merge queues we won't be able to prove who merged what from the git history, as that will be done by GitHub only (a single point).  But I suppose people won't see this as important, as
- it's GitHub who authorizes pushes either way, so I think it's "only" about retrospective auditing
- it's rare for people to sign the merges  (because it's a bit more work than pushing a button, I guess)

BTW, GitHub has a long history of commonly misattributing who merged which PR (and support politely refusing working on improving this) &ndash; at least for PRs merged by `git push`.  I don't recall seeing that issue in the past year (very roughly), at least.

wolfgangwalther: > I haven't thought through the tradeoffs fully. With mechanisms like merge queues we won't be able to prove who merged what from the git history, as that will be done by GitHub only (a single point).

If I understand correctly, the "I'll sign my merge commits" approach is mainly to prevent other committers from impersonating you - that's what you are protected against. You can say: Nope, this merge commit is not mine, because I always sign my commits.

Now.. this is only a problem, *if pushing directly to the target branch is possible in the first place*. If you can't push directly to the target branch, but let GitHub actually *create* those merge commits, then nobody can impersonate you anyway. At least I can't imagine how, because to impersonate you, I'd need to connect your email address to my account first.

Thus, I think, the case of "CLI merges are disallowed because we use merge queues", but "CLI merges are still *possible* by bypassing" is the worst case for you:
- You would not be able to consistently sign your commits anymore without restarting merge queues on every commit,
- But others are not stopped from sneaking in a commit pretending to be yours in a direct push, restarting the queue a single time, thus likely not being noticed.

If we strictly disallow CLI merges, this should make signing the merges not required anymore.

Does that make sense or what am I missing?

---

There's one aspect that I don't quite understand even right now, yet: We do have a requirement for pull requests to the default branch already. Thus, you can't merge arbitrary commits, they need to be in a pull request. When you now merge a PR via CLI, this merge will show up in the PR as well - with your name as the merger.

Now.. if you fake a merge commit from somebody else - will this *attribute the merge in the PR to this person as well?* I kind of doubt that this will happen. So I assume that in practice it's already impossible to fake a merge commit in another committers name right now, at least when targeting the default branch. (To require PRs and / or enable merge queues for other branches, we'd need to do https://github.com/NixOS/nixpkgs/issues/424345#issuecomment-3064099460)

Edit: I now see that this scenario seems to be precisely what you mentioned with "BTW, GitHub has a long history of commonly misattributing who merged which PR [...]". So this might need to be tested explicitly...

vcunat: Not really.  I just see a difference between GitHub promising that it checked stuff vs. having a permanent auditable cryptographic proof.

MattSturgeon: I sympathise with your position, but at a certain point we have to accept that we are using GitHub (warts and all) and may need to occasionally compromise by doing things their way. At least, if we want things like Merge Queues, which I believe most of us do.

> If I understand correctly, the "I'll sign my merge commits" approach is mainly to prevent other committers from impersonating you - that's what you are protected against. You can say: Nope, this merge commit is not mine, because I always sign my commits.

I would be very surprised if a bad merge being mis-atributed to you caused any serious damage to your reputation. Maybe I'm naive, but in that scenario you should be able to explain the known issue, reference this conversation, and prior examples of other mis-atributed merges.

---

Anecdotally, I've never _knowingly_ seen a mis-atributed merge commit created by GitHub; I'd be interested in seeing the examples you mentioned.

---

My personal view is that even if GitHub does do weird things on occasion, and we miss out on the ability to sign PR merge commits, the benefits of having a Merge Queue are still worth it.

wolfgangwalther: OK, with this input in mind, I changed the bypassers for the merge queue to all committers. Once we introduce the merge queue, everyone who merges via `git`'s CLI needs to be careful to only do this on non-merge-queue branches and not on master. For merges into master, the regular merge queue button in the PR must be used - otherwise the merge queue will be restarted on each CLI merge.

Note that this will be reported when pushing, too:

```
remote: Bypassed rule violations for refs/heads/master:
remote:
remote: - Changes must be made through the merge queue
remote:
```

It doesn't cause an error, though, so it might also just be missed.

vcunat: > I would be very surprised if a bad merge being mis-atributed to you caused any serious damage to your reputation.

To clarify, that wasn't the point at all.  And I didn't mean to argue for or against the queues, as I haven't thought about it much.  Just presenting explanations around the question sent at me.

wolfgangwalther:
This currently has conflicts. I will resolve those, once the following two issues are resolved:
- [x] Support merge queues in nixpkgs-merge-bot: https://github.com/NixOS/nixpkgs-merge-bot/issues/202
- [x] Merge https://github.com/NixOS/nixpkgs/pull/431146

rulesets/nixpkgs/require-merge-queue.json
wolfgangwalther: Turns out there is a simple solution to the problem of "accidentally bypassing the merge queue from the command line": The `bypass_mode` can be set to `pull_request` instead of `always`. We were not able to use that for required status checks, because these (currently) need to be bypassed for some of the branches when merging master back into them etc.

But that's not the case here. The merge queue only targets master for now anyway. I just tested, with the bypass mode to "pull request", it's not possible to do PR merges via `git` CLI anymore. Thus, there is also no risk to accidentally reset the queue when using the wrong workflow. It is still possible to bypass the queue via the checkbox in the PR, though - so in case it's needed, it's still there.

wolfgangwalther:
`nixpkgs-merge-bot` now supports merge queues, I rebased the branch and adjusted the bypass_mode to pull_request as mentioned in https://github.com/NixOS/org/pull/149#discussion_r2311988200.

This will enable the `lint` workflow for the merge queue. This workflow can currently fail for nixpkgs-vet and nixpkgs-vet mentions that in some special cases the PR should be merged anyway. This case is discussed in https://github.com/NixOS/nixpkgs-vet/issues/174. As long as that issue is not resolved, some PRs with nixpkgs-vet failures might need to be merged despite the required status checks failing (that's currently already the case).

I briefly wondered whether this means we should take nixpkgs-vet out of the merge queue for now or make it `continue-on-error: true` *for the merge queue only*, but that won't change a thing: Once the committer enables the "bypass" checkbox, *both* required status checks *and* merge queue will be bypassed anyway. Once the committer does, the merge queue check is not relevant anymore. Thus, there is no problem here.

---

From my perspective, this PR is ready to go: Let's enable the merge queue for master @NixOS/org!

rulesets/nixpkgs/require-merge-queue.json
MattSturgeon: > The `bypass_mode` can be set to `pull_request` instead of `always`.

Nice discovery!

> We were not able to use that for required status checks, because these (currently) need to be bypassed for some of the branches when merging master back into them etc.

I don't follow why that's a hard blocker.

Is it because this workflow avoids opening a PR against the protected branch at all, so we are bypassing the "require a pull request" rule? Or just that some prefer to bypass via `git push` instead of via GitHub's interface(s)?

> It is still possible to bypass the queue via the checkbox in the PR, though - so in case it's needed, it's still there.

And also via `gh pr merge --admin`, presumably.

wolfgangwalther: > > We were not able to use that for required status checks, because these (currently) need to be bypassed for some of the branches when merging master back into them etc.
> 
> I don't follow why that's a hard blocker.
> 
> Is it because this workflow avoids opening a PR against the protected branch at all, so we are bypassing the "require a pull request" rule? Or just that some prefer to bypass via `git push` instead of via GitHub's interface(s)?

Because merging master into staging creates a merge commit on staging. And this merge commit will *never* have any checks run without a PR. These branches don't have "require a pull request" set up, so far. But without a PR, they can never satisfy the required status checks, thus a manual `git push` always need to bypass the checks as well.

This applies for both automated and manual periodic merges.

Whether to move periodic merges to a PR-based workflow is discussed in #424345.

> > It is still possible to bypass the queue via the checkbox in the PR, though - so in case it's needed, it's still there.
> 
> And also via `gh pr merge --admin`, presumably.

Yes.

winterqt (approved):
Let's do it.

rulesets/nixpkgs/require-merge-queue.json
```diff
@@ -0,0 +1,36 @@
...
+        "max_entries_to_merge": 5,
+        "min_entries_to_merge_wait_minutes": 5,
+        "grouping_strategy": "ALLGREEN",
+        "check_response_timeout_minutes": 60
+      }
+    }
+  ],
+  "bypass_actors": [
+    {
+      "actor_id": 203427,
```
philiptaron, line 31:
```
$ gh api /user/203427
{
  "login": "mathieulh",
  "id": 203427,
  "node_id": "MDQ6VXNlcjIwMzQyNw==",
  "avatar_url": "https://avatars.githubusercontent.com/u/203427?v=4",
  "gravatar_id": "",
  "url": "https://api.github.com/users/mathieulh",
  "html_url": "https://github.com/mathieulh",
  "followers_url": "https://api.github.com/users/mathieulh/followers",
  "following_url": "https://api.github.com/users/mathieulh/following{/other_user}",
  "gists_url": "https://api.github.com/users/mathieulh/gists{/gist_id}",
  "starred_url": "https://api.github.com/users/mathieulh/starred{/owner}{/repo}",
  "subscriptions_url": "https://api.github.com/users/mathieulh/subscriptions",
  "organizations_url": "https://api.github.com/users/mathieulh/orgs",
  "repos_url": "https://api.github.com/users/mathieulh/repos",
  "events_url": "https://api.github.com/users/mathieulh/events{/privacy}",
  "received_events_url": "https://api.github.com/users/mathieulh/received_events",
  "type": "User",
  "user_view_type": "public",
  "site_admin": false,
  "name": "Mathieu HERVAIS (Mathieulh)",
  "company": null,
  "blog": "http://keybase.io/mathieulh",
  "location": "France",
  "email": null,
  "hireable": null,
  "bio": null,
  "twitter_username": null,
  "public_repos": 39,
  "public_gists": 2,
  "followers": 145,
  "following": 26,
  "created_at": "2010-02-14T12:16:56Z",
  "updated_at": "2024-12-20T09:45:16Z"
}
```

Is this right? Am I holding it wrong? How does one know what the actor id is?

`gh api /org/NixOS/teams` also doesn't show who this is, and `gh api /org/NixOS/teams/203427` shows 404.

Basically, I'm struggling to figure out how to say this number is correct.

wolfgangwalther: You will need to do:

```
gh api /orgs/NixOS/teams/nixpkgs-committers --jq '.id'
```

wolfgangwalther: Also note that these will be shown with name to whomever imports the rule. It's just the export that is ID based.

MattSturgeon: LGTM:

```
$ gh api /orgs/NixOS/teams/nixpkgs-committers --jq '.id'
203427
```

MattSturgeon (approved):

philiptaron (approved):

rulesets/nixpkgs/require-merge-queue.json
```diff
@@ -0,0 +1,36 @@
...
+    {
+      "type": "merge_queue",
+      "parameters": {
+        "merge_method": "MERGE",
+        "max_entries_to_build": 5,
+        "min_entries_to_merge": 1,
+        "max_entries_to_merge": 5,
+        "min_entries_to_merge_wait_minutes": 5,
+        "grouping_strategy": "ALLGREEN",
+        "check_response_timeout_minutes": 60
```
infinisil, line 25: Maybe check if these numbers make sense for Nixpkgs, or try them out and adjust as necessary (can do it more ad-hoc, commit later)

infinisil (approved):
LGTM, let's go!

rulesets/nixpkgs/require-merge-queue.json
wolfgangwalther: Yeah, no experience, so left the default here. We should certainly look into this once we have a bit more experience.

LunNova (approved):

winterqt:
Not merging while GitHub seems to be dying here in CH, but as soon as it's stable, let's do it.

wolfgangwalther:
The merge itself also won't do anything. This needs to be imported manually and be re-exported manually before merge.

winterqt:
I know! :)

winterqt:
This broke: https://github.com/NixOS/nixpkgs/pull/441598#event-19616977181

winterqt:
nvm

MattSturgeon:
> This broke: [NixOS/nixpkgs#441598 (comment)](https://github.com/NixOS/nixpkgs/pull/441598#event-19616977181)

Strange: that initial attempt didn't trigger the `merge-group.yml` workflow, but your second attempt [did](https://github.com/NixOS/nixpkgs/actions/runs/17620363876).

Looks like it's starting to work, I see three things have now been enqueued and each has a workflow run.

And as I finish typing this comment: it looks like everything went smoothly with the first few workflow runs!  🎉

winterqt:
Yes, I derped by copying the rule edit from this commit wrong and didn't notice till it was too late.

wolfgangwalther:
Please run the export script, too, to get the latest state of the rulesets back into the repo.

winterqt:
Already done: https://github.com/NixOS/org/pull/160

MattSturgeon:
_Maybe this doesn't fit well with technical discussion, and I don't want to distract anyone who's monitoring here for post-merge issues, but:_

I'd like to say **a big thank you to @wolfgangwalther** for all the effort they've put into getting this across the line.

Also thanks to @winterqt for applying the changes to the ruleset and monitoring for issues.

And finally, thanks to everyone else for their reviews, feedback, etc.

_Hopefully this continues to go smoothly, but if not I have every faith we will identify whatever issues exist and iron them out._

wolfgangwalther:
A few notes from looking at things:
- The current merge queue for each branch can be seen in the branches tab. For master it leads to https://github.com/NixOS/nixpkgs/queue/master.
- I have not seen a nixpkgs-merge-bot merge, yet. Will want to confirm this works, too. Works: https://github.com/NixOS/nixpkgs/pull/441717

PR #155, opened 2025-08-31
rulesets/nixpkgs: disallow squash/rebase merges on development branches

### LINKED ISSUE #135: nixpkgs: disable rebase and squash merges
While [working through the contribution guidelines](https://github.com/NixOS/nixpkgs/pull/421104), we came across this:

> Pull requests should not be squash merged in order to keep complete commit messages and GPG signatures intact and must not be when the change doesn't make sense as a single commit.

We wondered whether we should just outright forbid squash merges.

Are there any scenarios where squash merge is *required*?

Most cases could probably be worked around by pushing to the contributors branch directly. Maybe a small how to could help committers to set things up for that.

Assuming we don't require squash merges, I'd thus like to propose to disable them (and rebase merges, too)

wolfgangwalther:
With the [introduction of Merge Queues](https://github.com/NixOS/org/pull/149), the only merge method on those branches will be "merge" as configured for the Merge Queue. By disallowing the other merge methods on development branches without merge queue, we [achieve consistency without relying on contributor documentation](https://github.com/NixOS/nixpkgs/pull/438686#discussion_r2312393859).

Resolves #135

### DIFF
--- a/rulesets/nixpkgs/require-merge-method.json
+++ b/rulesets/nixpkgs/require-merge-method.json
@@ -1,0 +1,41 @@
+{
+  "name": "require-merge-method",
+  "target": "branch",
+  "source_type": "Repository",
+  "source": "NixOS/nixpkgs",
+  "enforcement": "active",
+  "conditions": {
+    "ref_name": {
+      "exclude": [],
+      "include": [
+        "~DEFAULT_BRANCH",
+        "refs/heads/release*",
+        "refs/heads/staging*",
+        "refs/heads/haskell-updates"
+      ]
+    }
+  },
+  "rules": [
+    {
+      "type": "pull_request",
+      "parameters": {
+        "allowed_merge_methods": [
+          "merge"
+        ]
+      }
+    }
+  ],
+  "bypass_actors": [
+    {
+      "actor_id": 203427,
+      "actor_type": "Team",
+      "bypass_mode": "always"
+    },
+    {
+      "actor_id": 1075715,
+      "actor_type": "Integration",
+      "bypass_mode": "always"
+    }
+  ],
+  "current_user_can_bypass": "always"
+}


rulesets/nixpkgs/require-merge-method.json
```diff
@@ -0,0 +1,41 @@
...
+      "actor_id": 203427,
+      "actor_type": "Team",
+      "bypass_mode": "always"
+    },
+    {
+      "actor_id": 1075715,
+      "actor_type": "Integration",
+      "bypass_mode": "always"
+    }
+  ],
```
wolfgangwalther, line 39: Committers and nixpkgs-ci are bypassers here. Not for the merge methods, but for the "require pull request" part, we still need to be allowed to push to branches for the periodic merges.

While this will allow bypassing a PR entirely, it will not allow to use any other buttons than "merge". Squash and Rebase are gone, I tested that.

Note that PRs are still required for master/release branches via `require-pull-request-except-release-team.json`.

wolfgangwalther:
Could someone re-run CI here? The references check is slightly annoying.

philiptaron:
I can see on the [NixOS Infra Alerts](https://matrix.to/#/#infra-alerts:nixos.org) matrix channel that the Wiki does in fact appear to be going up and down.

MattSturgeon (approved):

wolfgangwalther:
I tried implementing this, but it doesn't work. I'm not sure what I did differently while testing it in my fork, but I think I must have made a mistake while testing. I can't find a combination of "allowing pushes" and "forcing merge methods in the UI".

This could only be implemented once pushes are forbidden. But once that's the case, we'd enable a Merge Queue for these branches anyway, so this becomes implied.

Nothing we can do here, I think :/

PR #171, opened 2025-09-21
rulesets/nixpkgs: switch merge queue to "headgreen" strategy

wolfgangwalther:
The `ALLGREEN` setting corresponds to "Require all queue entries to pass required checks". `HEADGREEN` is the opposite, when disabling that checkbox, which then behaves like this:

> only the commit at the head of the merge group, i.e. the commit
> containing changes from all of the PRs in the group, must pass its
> required checks to merge.

This should fix an odd scheduling behavior we're seeing for a few days now: With a reasonable number of jobs queued, GitHub will sometimes just forget to schedule jobs for a merge queue item. These will then wait for an hour for the status check to timeout, be kicked out of the queue and all other jobs run again.

We don't really need to do that, though: We already require status checks for each PR individually, thus there is a *very low chance* of introducing introducing a commit that:
- Passes CI in the PR, and
- Fails CI in the Merge Queue, and
- Is fixed by a random PR queue up behind it.

While at it, also reduce the timeout from 60 minutes to 20 minutes, because we'll certainly not need more time for any merge to master, even if we had the Eval and Build workflows running (which is not even the case, yet).

---

The "broken" state of the merge queue looks like this:

2025-09-21T15:14:32,696351609+02:00

Note how the first item has been queued for more than 20 minutes and still didn't post a result. Going into the Actions tab, we can also see that GitHub never scheduled a run for this commit...

cc @NixOS/nixpkgs-ci

### DIFF
--- a/rulesets/nixpkgs/require-merge-queue.json
+++ b/rulesets/nixpkgs/require-merge-queue.json
@@ -22,8 +22,8 @@
         "min_entries_to_merge": 1,
         "max_entries_to_merge": 5,
         "min_entries_to_merge_wait_minutes": 5,
-        "grouping_strategy": "ALLGREEN",
-        "check_response_timeout_minutes": 60
+        "grouping_strategy": "HEADGREEN",
+        "check_response_timeout_minutes": 20
       }
     }
   ],


rulesets/nixpkgs/require-merge-queue.json
```diff
@@ -22,8 +22,8 @@
         "min_entries_to_merge": 1,
         "max_entries_to_merge": 5,
         "min_entries_to_merge_wait_minutes": 5,
```
MattSturgeon, line 24: While we're talking about finessing the numbers, we should look at tweaking these numbers soon.

In particular, I think we should:
- Increase "max entries" (build & merge), so that we test & merge larger groups
- Increase "min entries," so that we don't start testing until we have a few PRs queued
- Consider decreasing "minutes to wait," so that we don't have to wait too long if the queue is really quiet for some reason (maybe 5m is ok already, though?)

Increasing the "max entries" is based on the assumption that failures will be rare. Therefore, we should optimistically attempt to test & merge larger groups, thereby speeding up the queue. Failures will delay more PRs, but the happy path will be faster.

MattSturgeon (approved):
Seems good 👍

rulesets/nixpkgs/require-merge-queue.json
wolfgangwalther: Before tweaking these numbers, I think we should move more stuff to the queue first. Eval, Build etc. - right now, with the very basic queue, it might have very different behavior.

Mic92 (approved):

infinisil (approved):
Sounds good, let's try it out

Lassulus (approved):

infinisil:
Applied!

wolfgangwalther:
Thank you!

PR #172, opened 2025-09-21
rulesets: add batch-with-kernel-updates

Ma27:
See https://github.com/NixOS/nixpkgs/issues/437208#issuecomment-3254242035 for context.

cc @wolfgangwalther @k900 @alyssais

### DIFF
--- a/rulesets/nixpkgs/no-creation.json
+++ b/rulesets/nixpkgs/no-creation.json
@@ -18,7 +18,8 @@
         "refs/heads/wip-*/**/*",
         "refs/heads/revert-*/**/*",
         "refs/heads/dependabot/**/*",
-        "refs/heads/gh-readonly-queue/**/*"
+        "refs/heads/gh-readonly-queue/**/*",
+        "refs/heads/batch-with-kernel-updates"
       ],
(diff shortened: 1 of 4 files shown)


wolfgangwalther (changes requested):
This branch should also be added to the `no-force-push` ruleset.

Also, I think you will need to add it to `no-creation` as well, otherwise... you can't create it in the first place.

wolfgangwalther (approved):
I think this should work, yes.

wolfgangwalther:
I discussed the name a lot with @emilazy in https://github.com/NixOS/nixpkgs/issues/437208. While I believe `kernel-updates` is better than `staging-nixos`, I also believe that both are way better than `batch-with-kernel-updates`. The former two follow already established patterns.

Thus, I agreed with @Ma27 on Matrix to go with `staging-nixos`.

I checked - and all branch protection rules already support this branch name accordingly. Thus I only temporarily disable the no-creation-except-release-team rule to allow myself to create the branch, then enabled it again.

The `staging-nixos` branch now exists, and I'm closing this as "completed".

PR #181, opened 2025-10-04
constitution: count vacant seats when voting full reelection

wolfgangwalther:
The way the constitution is currently written, the SC can entirely deadlock, if one seat is vacant and 50% of the remaining SC are calling for no confidence, while the other 50% disagree.

The SC can neither be fully re-elected, nor make any other decisions with simple majority.

<details>
<summary>I also considered a few alternatives, which I have not taken for various reasons.</summary>

## 1. Flip to "vote of confidence"

Change the above to:

> Any SC member may call for a vote of confidence. If this vote does not pass with a simple majority, the entire SC will be reelected based on perceived loss of confidence.

Pro:
- Resolves the deadlock.

Contra:
- Creates a "pending" state for regular case without vacant seats, probably would need a clear deadline for votes.
- Unclear what happens with special vs initial election.

## 2. Pass on ties

Change the above to:

> A simple majority within the SC may call a reelection of the entire SC based on perceived loss of confidence. A tie counts as majority.

Pro:

Contra:
- Doesn't solve the deadlock, because 4 votes are needed anyway, so *there are no ties*.
- Unclear what happens with special vs initial election.

## 3. Count vacant seats as reelection only

Change the above to:

> A simple majority within the SC may call a reelection of the entire SC based on perceived loss of confidence. Vacant seats count as voting for reelection.

Pro:
- Resolves the deadlock.

Contra:
- Unclear what happens with special vs initial election.
</details>

I thank @7c6f434c for helping with the wording here.

### DIFF
--- a/doc/constitution.md
+++ b/doc/constitution.md
@@ -108,7 +108,7 @@ Deciding that an amendment is not substantial can be done by unanimity among a f
 
 #### Ordinary decisions
 
-Everything within the authority of SC that doesn't require an extraordinary decision can be decided with a simple majority (more than 50%) of a full SC (vacant seats count as abstaining, except when voting for reelection).
+Everything within the authority of SC that doesn't require an extraordinary decision can be decided with a simple majority (more than 50%) of a full SC (vacant seats count as abstaining, [except when voting for reelection](#full-reelections)).
 
 An exception to that is when there is when the [conflict of interest balance](#conflict-of-interest-coi-balance) condition is suspected to be violated.
 In that case, all non-involved currently serving SC members will have a simple majority decision over whether there is a violation.
--- a/doc/constitution.md
+++ b/doc/constitution.md
@@ -108,7 +108,7 @@ Deciding that an amendment is not substantial can be done by unanimity among a f
 
 #### Ordinary decisions
 
-Everything within the authority of SC that doesn't require an extraordinary decision can be decided with a simple majority (more than 50%) of a full SC (vacant seats count as abstaining).
+Everything within the authority of SC that doesn't require an extraordinary decision can be decided with a simple majority (more than 50%) of a full SC (vacant seats count as abstaining, except when voting for reelection).
 
 An exception to that is when there is when the [conflict of interest balance](#conflict-of-interest-coi-balance) condition is suspected to be violated.
 In that case, all non-involved currently serving SC members will have a simple majority decision over whether there is a violation.
@@ -205,6 +205,7 @@ A committee member elected in a special election will serve out the remainder of
 
 A simple majority within the SC may call a reelection of the entire SC based on perceived loss of confidence.
 In this case, it also has to be decided whether this election is considered a special election for the remainders of all the corresponding terms, or an initial election for full 2-year terms for half of the seats rounded up and 1-year half-terms for the remaining seats.
+Vacant seats vote in favour of reelection, and between initial election and special election they count towards special election.
 
 ### Removal for conduct
 


JulienMalka (approved):

nim65s (approved):
Thanks a lot for this, I objectively think this is a very important task and it certainly did require a lot of work.

(More subjectively, I'm currently running, so I think I should only be supporting the work of an independent party, and not be pushing on this direction myself)

Lassulus (approved):
I think this more matches the intention we wanted to have when designing this originally

niklaskorz (approved):

rhendric:
If there are two vacant or otherwise indisposed seats, and of the five sitting members, two vote for a re-election and three against, with this amendment, the re-election would pass. Do we want that outcome?

On the one hand: perhaps an SC that is running at diminished capacity should be more sensitive to finding internal consensus, or more motivated to hold a special election to get back to full size.

On the other hand: when a minority feels out of power, this would incentivize them to flip the table and take their chances at getting in power, and the community would pay the cost of more frequent re-elections. I don't think we want to normalize that.

JulienMalka:
All majority decisions in the SC require 50% of the full SC, vacant seats counting as abstention. So if there is 2 vacant seats, a majority decision still requires 4 positive votes. Meaning 2 people being in favor of disbanding the SC can block all decisions, they might as well be able to trigger a no-confidence vote. 

I like that this proposal also put clear incentives on replacing vacant seats quickly.

rhendric:
> All majority decisions in the SC require 50% of the full SC, vacant seats counting as abstention. So if there is 2 vacant seats, a majority decision still requires 4 positive votes. Meaning 2 people being in favor of disbanding the SC can block all decisions, they might as well be able to trigger a no-confidence vote.

Flawless argument, thank you. That gets me off the fence.

rhendric (approved):

pluiedev (approved):

doc/constitution.md
```diff
@@ -99,7 +99,7 @@ Substantial amendments to the Nix Community Values require 90% agreement in a po
 
 #### Ordinary decisions
 
-Everything within the authority of SC that doesn't require an extraordinary decision can be decided with a simple majority (more than 50%) of a full SC (vacant seats count as abstaining).
+Everything within the authority of SC that doesn't require an extraordinary decision can be decided with a simple majority (more than 50%) of a full SC (vacant seats count as abstaining, except when voting for reelection).
```
cafkafk: The current phrasing creates an ambiguity regarding reelection votes. 

It explicitly states that vacant seats do not count as `abstaining` but it fails to specify how they should be treated. Making explicit that they're not counted for or against, but are in fact not counted at all may be a sensible clarification.

```suggestion
Everything within the authority of SC that doesn't require an extraordinary decision can be decided with a simple majority (more than 50%) of a full SC (vacant seats count as abstaining, except when voting for reelection, in which case they are excluded from the total).
```

emilazy (approved):

cafkafk (changes requested):
I think this is an important clarification, after that I'm completely on board with the intent here.

Thanks for also clarifying what other options you considered!

doc/constitution.md
wolfgangwalther: That's not the case, though. They *are* counted. They are counted as "voting in favor of reelection" as the other changed paragraph further down specifies.

philiptaron (approved):
Being deadlocked for a long period of time is destabilizing to the community and to governance. This proposed amendment provides the appropriate finesse to recognize that instability without creating even more. Kudos to @wolfgangwalther, @JulienMalka, and @7c6f434c for crafting it.

drupol (approved):

doc/constitution.md
cafkafk: Ahh, then all the more reason to specify it, since that wasn't my initial reading of the paragraph.

Ideally, with a reference to the section where the rules are specified. A less intrusive solution:
```suggestion
Everything within the authority of SC that doesn't require an extraordinary decision can be decided with a simple majority (more than 50%) of a full SC (vacant seats count as abstaining, [except when voting for reelection](#full-reelections)).
```

That would be harmonious with how [the next line](https://github.com/wolfgangwalther/nixos-org/blob/7f777ac3fd36e734d0faaa88a27fd99eea17a82d/doc/constitution.md?plain=1#L104) references special cases.

A stronger version harmonious with the entire sections may be to change this hunk to a line similar to [L104](https://github.com/wolfgangwalther/nixos-org/blob/7f777ac3fd36e734d0faaa88a27fd99eea17a82d/doc/constitution.md?plain=1#L104) for the case of a reelection, but I can see why making the changes as unobtrusive as possible may be favorable.

jopejoe1 (approved):

doc/constitution.md
wolfgangwalther: I added the cross-reference, which should be uncontroversial enough to not require everyone to approve again. As you say it's also the less invasive change.

cafkafk (approved):

Gabriella439 (approved):

winterqt (approved):

tomodachi94 (approved):
Yes, let's do this please. (I don't really have anything else to say here other than what has already been said.)

mschwaig:
Making the no-confidence mechanism more powerful clashes with the idea of having staggered terms. The no-confidence vote could be used by SC members who's term is running out soon to make more seats available for re-election.

We have to take (perverse) incentives like that into account when we consider measures to make the SC more robust.

mschwaig:
Another thing we could take into account with the no-confidence vote if we wanted to is for how long the has been in a 'hung' state, and how far away the next regular election is.

---

IMO it is a good thing if we can make it through a full term with a healthy SC most of the time, and we should design for that.

Addressing the diminished capacity of the SC based on the outcome of the previous election might with help with keeping the no-confidence mechanism reachable, and keep the SC full.

lilyball:
> Making the no-confidence mechanism more powerful clashes with the idea of having staggered terms. The no-confidence vote could be used by SC members who's term is running out soon to make more seats available for re-election.

You’re suggesting that SC members would disband the entire SC to make more room so that way they’re more likely to win a reelection? Why would anyone vote for an SC member who behaves that way?

KiaraGrouwstra (approved):
This is definitely a step in the right direction.

In fact, I would personally go further still, and have vacant seats vote not in favor of special election, but in the case a tie-breaker is needed have them in fact vote in favor of a full reelection, similar in spirit to #472.

Now, given this proposal would make it easier to 'un-stagger' the delegates, I should probably address that topic as well:

> Making the no-confidence mechanism more powerful clashes with the idea of having staggered terms. The no-confidence vote could be used by SC members who's term is running out soon to make more seats available for re-election.

I think this begs the question of the intent behind staggered staggered elections, and I could think of two:

1. offering continuity between SC terms
1. obstructing hostile take-overs

Now, I presume offering continuity was the original intent, and I believe full re-elections aren't necessarily a crucial concern for this - popular incumbent candidates may simply be re-elected.

Obstructing hostile take-overs, which I believe @mschwaig is getting at here, I don't really understand so much as a concern: I believe the current commit threshold to become eligible as a voter already makes it somewhat unlikely an entity would just manage to obtain a significant number of sock puppets.

In fact, I believe erring (if only to the extent of a tie-breaker) in favor of full re-election here, if anything, would mirror the intent expressed at the current [call for re-election](https://discourse.nixos.org/t/call-for-full-re-election-of-the-steering-committee/70208).

In general, I believe the concept of filling separate seats runs counter to the spirit of proportional representation: if seats are filled one by one, the result is the majority position would win each of those seats, effectively defeating the representation of minority positions that proportional representation would protect.

jtojnar (approved):

alyssais (approved):

7c6f434c:
> if seats are filled one by one

A special election for _all_ seats mean a single multi-seat special election; the difference with new-initial election is that the terms correspond to the remainders of the (staggered) terms, not to new 1 and 2 year terms from the new date.

> You’re suggesting that SC members would disband the entire SC to make more room so that way they’re more likely to win a reelection? Why would anyone vote for an SC member who behaves that way?

Given how proportionality works, for representatives of cohesive minorities below 20% but above 13% of active voters (so, a high chance of SC seat in full reelection and a low chance of SC seat in a regular one, even at 4 seats) this might in some sense be the best representation of their constituents' interests!

MultisampledNight (approved):

totoroot (approved):

psionic-k:
Biasing towards full re-election through a no-confidence mechanism biases towards more instability and risk.  IMO the "bug" is here:

> a special election for that position may be held if the SC deems it necessary, or the SC does not have half of the normal size

Change this to:

> a special election for that position **must** be held **unless the SC can decide against it** or the SC does not have half of the normal size

Reason to favor special election mechanism instead of no-confidence mechanism:
- retains more working knowledge of the current SC members and defaults towards continuity
- biases towards seat representing the voting majority, which is more likely to produce a stronger consensus
- results in a less pivotal non-standard election timing, meaning fewer people will miss it due to it not being the normal timing
- voters can just hurriedly pick one candidate instead of needing to carefully consider several
- the no-confidence mechanism of resolving empty seats and resulting deadlocks still requires no-confidence votes, biasing towards drama

LunNova (approved):

Ericson2314:
This constitutional "bug" (as some of you have called it) is something that if either "fixed" or left in place, will benefit or dis-benefit some side of the conflict. I think everyone can agree that this is a simple fact.

A longstanding personal view of mine, since way before this, that the (real life) law is full of bugs and "tech debt", and there should be more refactoring to fix it --- just like I like to do with open source projects. But I also recognized that wasn't going to happen in legislatures with very high interpersonal tensions, where everything is viewed through the lens of a "win" or a "loss".

I therefore think we should all agree to drop the "bug" framing here.

Finally, as a practical matter, we became less deadlocked in the last meeting (notes at: https://discourse.nixos.org/t/sc-meeting-2025-10-05-21-00-utc/70418). I hope this trend will continue, and so this constitutional quandary becomes less urgent. If it does, we *will* then be able to approach it more apolitically, after all.

-----

On a separate note, a few people in this thread have raised other issues, such as the incentive for termed-out people to also remove their opponents, and the advantages of always quickly restoring the SC to its full size to avoid this situation in the first place. At the very least, that should remind us that to the extent there is a problem, this is not the only possible solution, and not a solution that doesn't come with its own set of drawbacks.

Zimmi48 (approved):

patka-123 (approved):

SigmaSquadron (approved):

piegamesde:
@Ericson2314 so what you're saying is "because the current situation benefits my 'side' by virtue of holding me in power, actually this is not a bug but a feature". What a stellar role model of leadership putting personal interests over what's best for the community.

(To be fair I don't think this PR matters too much in the grand scheme of things, it would a small drop in a large bucket given how fucked things are, but I can't let this stand without pointing out that this of line of argumentation is bad optics at best and bad leadership at worst.)

vcunat:
Pie: I believe it's hard to be objective at this moment.  The author and lots of endorsers of this PR want to get the current SC to end, and it doesn't appear that this tiny perhaps-bug is the main reason they want the SC to end now.

It's therefore very unfortunate to discuss this change at a point where there's a clear power struggle where this PR would significantly shift the balance at this very moment.  But I understand it happens; things often don't catch people's interest until they matter a lot.

piegamesde:
@vcunat yes I'm aware, and to be clear I'm not trying to advocate for merging this proposal now or whatever. I just wanted to voice my disagreement with John's line of argument for not merging it, because I find it misguided.

Ericson2314:
> @Ericson2314 so what you're saying is "because the current situation benefits my 'side' by virtue of holding me in power, actually this is not a bug but a feature"

Just so everyone is clear, I don't think that is an accurate paraphrase of what I wrote. It certainly isn't what I meant to convey. The meaning I intended is basically what @vcunat wrote.

philiptaron (approved):
I remain in favor of this amendment.

doc/constitution.md
```diff
@@ -178,7 +178,7 @@ In the event of a resignation or other loss of a steering committee member (incl
 
 #### Full Reelections
 
-A simple majority within the SC may call a reelection of the entire SC based on perceived loss of confidence. In this case, it also has to be decided whether this election is considered a special election for the remainders of all the corresponding terms, or an initial election for full 2-year terms for half of the seats rounded up and 1-year half-terms for the remaining seats.
+A simple majority within the SC may call a reelection of the entire SC based on perceived loss of confidence. In this case, it also has to be decided whether this election is considered a special election for the remainders of all the corresponding terms, or an initial election for full 2-year terms for half of the seats rounded up and 1-year half-terms for the remaining seats. Vacant seats vote in favour of reelection, and between initial election and special election they count towards special election.
```
Ericson2314: I would like to include @JulienMalka's excellent and very subtle note as some sort of official, but extra legal commentary. I think it is important to include in order to guide the readers / "judges" of future disputes, and prevents the original intent from being lost (GitHub comments are less durable than in-repo text).

(This can be analogized to commentaries and precedent in Common Law systems, but hopefully way less jank and fuzzy.)

This could take the form of a block quote like the following:

```suggestion
A simple majority within the SC may call a reelection of the entire SC based on perceived loss of confidence. In this case, it also has to be decided whether this election is considered a special election for the remainders of all the corresponding terms, or an initial election for full 2-year terms for half of the seats rounded up and 1-year half-terms for the remaining seats. Vacant seats vote in favour of reelection, and between initial election and special election they count towards special election.

> **Commentary**
>
> It might seem instead simpler to change the vote for full reelection (in cases of no confidence) to a vote for no reelection (in cases of confidence).
> This inversion would change the ">= 50%" threshold back to "> 50%", matching the votes, seemingly removing the need for a special case in the voting rules.
> The problem is that this also changes the interpretation of member who haven't yet or won't vote at all.
> Without introducing another special case, confidence is lost until the enough members actively reaffirm it by voting "yes".
> To keep the ">= 50%" threshold for votes of favor of reelection, while maintaining the burden of action on the voters to reelect (confidence is the default, the loss of confidence requires explicit action), special cases would thus be needed either way.
```

wolfgangwalther: I don't think there is a need to reinvent a VCS or forge. People can `git blame` the constitution and look at the PR for further discussion.

Lassulus: it could also be included in the commit message? so it would not be tied to the forge but would not clutter the constitution?

wolfgangwalther: Amended the commit message accordingly, did not make any changes to the file.

Ericson2314: I would not call this "reinventing VCS". If this was code, this is the sort of information I would put in a code comment, rather than a commit message. That is because it describes the code/normative part, not the change. (By contrast, the commit messag might indicate the recent history of 3-3 no-confidence votes.)

Ericson2314: A bit closer to home, W3C specs (e.g. https://www.w3.org/TR/SVG2/conform.html#secure-animated-mode) have long included block code commentary like that. I've always enjoyed reading them whenever I need to consult such a spec.

KiaraGrouwstra: language nits:
- the enough -> enough
- of favor of -> in favor of

KiaraGrouwstra:
@wolfgangwalther:
> Creates a "pending" state for regular case without vacant seats, probably would need a clear deadline for votes.

Could you elaborate on the issue here? Is this not resolved simply by considering the matter ready to call the result on once enough votes have been cast to settle the matter?

(To be fair, this reading interprets 'vacant seat' as an official state that may be formally distinguished from absenteeism. I noticed this term has been been used already - perhaps an NCA member could clarify its intended meaning?)

7c6f434c:
Vacant seat in a committee is a seat whose holder has resigned or has been removed, and thus there is no person who can officially write into the discussion channel with a vote pro or contra. (The SC is likely to run some votes asynchronously, and in such votes absentee members who have not resigned — nor have been absentee enough for others to remove officially — could vote without showing up to any sync or semi-sync discussions)

doc/constitution.md
wolfgangwalther: I removed the comment from my commit message again. I have no intention to copy-edit this stuff or receive nit picks about it.

wolfgangwalther:
I'll keep this PR exactly as is. If the SC decides anything different from this PR, they can easily put up their own PR. I have zero interest to fight the Nix team in stalling things again and again.

wolfgangwalther:
I noticed this is merge conflicted now (how likely is that on a constitutional change!), but "Allow edits by maintainers" is turned on, so the SC can easily fix the conflicts if they decide to vote for this _as-is_.

roberth (approved):
I approve this change.

@wolfgangwalther, thank you for proposing it, and I apologize for having to delay this.

I have not formally approved it earlier, because of the highly politicized environment in which this proposal and a related proposal came about.

Specifically, it is *not OK* to for a Steering Committee member to engage in a risky effort and then pull the plug when things don't go quite they hoped. That is the time to follow through and take responsibility for your actions; not to take the community hostage to try and save their own political ass [reputation]. That is unfortunately what happened.
Under those circumstances I could of course not give in to their demands, and approving this earlier would have amounted to doing so.

Someone could try and accuse me of something similar, but their argument would fall flat on their face. I have let my reputation take a hit by remaining part of the "establishment" that wanted to fix things. I have lived through a pressure campaign. I've been unproductive in my work. All while I could have just said: fuck it, let them have their way and I'll do my own things in peace now. And I wished I did.

But then I would have screwed over all future Steering Committees and the community they represent, by showing that political games and pressure campaigns would still work.

That is not how we do things anymore.

SC members *will* resist manipulation, and pressure campaigns only serve to make the community worse for everyone.
If your idea is good, it *will* be passed, *through the established governance processes*, and that includes this change.
Sounds uncool, but how we have to be serious about how we handle disagreements, and continuous public outrage is just not necessary and not conducive to a healthy community that people want to be involved with or that they can in any way depend on in their careers and lives.

I want an _effective_ SC that works with the community for the community. I have previously voted to have a standing Election Committee to ensure the SC is restaffed without hesitation, and this amendment provides a further backstop to avoid deadlock if all else fails.

Thanks again, and feel free to make proposals and engage in dialogue here, in other places, or to reach out to SC members or the [whole SC](https://nixos.org/community/teams/steering-committee/) as well.

Ericson2314:
https://github.com/NixOS/org/compare/ff7f2ea0069a80058a9aa95ee96a6ffa688b874b..372a2a4b3406c0a382dd2ffd7201717b358dc05f?short_path=267c4ae#diff-267c4aea889578871ea0344028a990278abfa91dacee79dac88f07ff7bb59c23 and "rich diff" verifies the rebase did not change the meaning

```
git diff ff7f2ea0069a80058a9aa95ee96a6ffa688b874b..372a2a4b3406c0a382dd2ffd7201717b358dc05f --ignore-all-space --word-diff doc/constitution.md
```
also does that locally

infinisil:
I see that 4 SC members (@JulienMalka @roberth @philiptaron @cafkafk) have approved this PR, but we require 5 approvals for constitutional amendments. @Ericson2314 Can you confirm that you or somebody else has given the 5th approval?

JulienMalka:
There is a 5th approval. We need to formally announce it but there is a vote log, too: https://github.com/NixOS/steering-committee/blob/main/vote-logs/2025-2026.md

PR #183, opened 2025-10-12
rulesets/nixpkgs: enable merge queue on secondary development branches

### LINKED ISSUE #135: nixpkgs: disable rebase and squash merges

wolfgangwalther:
This enables Merge Queues on secondary development branches like staging and haskell-updates as well.

These branches still need to be pushed to manually for the periodic merges, both automated by CI and manually by committers. Compared to the [ruleset for merge queues on the primary development branches](https://github.com/NixOS/org/blob/main/rulesets/nixpkgs/require-merge-queue.json), this `always` allows bypassing. Since the number of merges on these branches is much lower, an occasional bypassing of the queue for the periodic merges is not a problem, at least resource-wise.

We will still want to eventually implement a [PR-based workflow for periodic merges](https://github.com/NixOS/nixpkgs/issues/424345) to make sure these never break the branch.

Although not the primary motivation, this will also resolve #135.

cc @NixOS/nixpkgs-ci

### DIFF
--- a/rulesets/nixpkgs/require-merge-queue-except-periodic-merges.json
+++ b/rulesets/nixpkgs/require-merge-queue-except-periodic-merges.json
@@ -1,0 +1,49 @@
+{
+  "id": 8822191,
+  "name": "require-merge-queue-except-periodic-merges",
+  "target": "branch",
+  "source_type": "Repository",
+  "source": "NixOS/nixpkgs",
+  "enforcement": "active",
+  "conditions": {
+    "ref_name": {
+      "exclude": [],
+      "include": [
+        "refs/heads/haskell-updates",
+        "refs/heads/staging",
+        "refs/heads/staging-next",
+        "refs/heads/staging-25.05",
+        "refs/heads/staging-next-25.05"
+      ]
+    }
+  },
+  "rules": [
+    {
+      "type": "merge_queue",
+      "parameters": {
+        "merge_method": "MERGE",
+        "max_entries_to_build": 5,
+        "min_entries_to_merge": 1,
+        "max_entries_to_merge": 5,
+        "min_entries_to_merge_wait_minutes": 5,
+        "grouping_strategy": "HEADGREEN",
+        "check_response_timeout_minutes": 20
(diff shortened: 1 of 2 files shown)


MattSturgeon (approved):

wolfgangwalther:
Thanks for the quick look.

This is already implemented, but merge is blocked because I fixed the `export.bash` script. @infinisil can you have a look and approve that change?

K900 (approved):

infinisil (approved):

wolfgangwalther:
Thanks!

rulesets/nixpkgs/require-merge-queue-except-periodic-merges.json
```diff
@@ -0,0 +1,49 @@
...
+  "source_type": "Repository",
+  "source": "NixOS/nixpkgs",
+  "enforcement": "active",
+  "conditions": {
+    "ref_name": {
+      "exclude": [],
+      "include": [
+        "refs/heads/haskell-updates",
+        "refs/heads/staging",
+        "refs/heads/staging-next",
```
zowoq, line 14: Should `staging-nixos` be added here?

wolfgangwalther: Yes. Could you open a PR?

zowoq: https://github.com/NixOS/org/pull/201

PR #186, opened 2025-10-28
rulesets/nixpkgs: Allow create-pull-request prefix branches

infinisil:
For workflows that create Nixpkgs pull requests. See https://github.com/NixOS/nixpkgs/pull/450864#discussion_r2467509174

### DIFF
--- a/rulesets/nixpkgs/no-creation.json
+++ b/rulesets/nixpkgs/no-creation.json
@@ -18,7 +18,8 @@
         "refs/heads/wip-*/**/*",
         "refs/heads/revert-*/**/*",
         "refs/heads/dependabot/**/*",
-        "refs/heads/gh-readonly-queue/**/*"
+        "refs/heads/gh-readonly-queue/**/*",
+        "refs/heads/pr/**/*"
       ],
       "include": [
         "~ALL"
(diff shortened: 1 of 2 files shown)


rulesets/nixpkgs/no-creation.json
```diff
@@ -18,7 +18,8 @@
         "refs/heads/wip-*/**/*",
         "refs/heads/revert-*/**/*",
         "refs/heads/dependabot/**/*",
-        "refs/heads/gh-readonly-queue/**/*"
+        "refs/heads/gh-readonly-queue/**/*",
+        "refs/heads/create-pull-request/**/*"
```
wolfgangwalther: We might use this namespace for periodic merges, too, if we move them to PRs. So "PR" in general seems good to me.

Our own names are much shorter in general. (the `gh-readonly-queue` thing is given by GitHub).

I'd prefer something like `pr-` or `pr/**/*`, which has the same meaning.

wolfgangwalther (commented):
To avoid accidental collisions, we should limit this namespace to *automation*, not regular committers. Regular committers can use `wip-`. I added a separate rule to do this.

wolfgangwalther (approved):
I implemented this with my own feedback addressed.

rulesets/nixpkgs/no-creation.json
```diff
@@ -30,8 +31,8 @@
       "type": "creation"
     }
   ],
-  "created_at": "2025-06-11T14:44:19.501-04:00",
-  "updated_at": "2025-09-10T12:29:36.746-04:00",
+  "created_at": "2025-06-11T20:44:19.501+02:00",
+  "updated_at": "2025-10-28T09:52:46.205+01:00",
```
wolfgangwalther, line 35: Note: The diff for `created_at` is stupid. We should make this script do UTC instead - or remove the `created_at` / `updated_at` fields entirely.

PR #187, opened 2025-10-28
rulesets/export: remove timestamp

wolfgangwalther:
No need to track timestamps - we have the git log for that.

Related: https://github.com/NixOS/org/pull/186/files#r2468561407


infinisil (approved):
Nice!

emilazy (approved):

PR #197, opened 2025-11-06
Reformat text to have at most one sentence per line

Ericson2314:
The reason for this is given in the new section of the read-me.

### DIFF
--- a/doc/constitution.md
+++ b/doc/constitution.md
@@ -32,7 +32,8 @@ It has the following responsibilities, which it may delegate as appropriate:
     - Decision authority may be delegated to other teams when significant analysis is needed, but the SC remains responsible if the delegated team fails to reach a decision.
   - Evolve the constitution, including governance structures and processes, as needed, subject to supermajority approval.
 - Coordination with the NixOS Foundation Board:
-  - Collaborate with the NixOS Foundation Board to set policies and priorities for high-visibility actions, events, and resources under the Foundation's care. Both the SC and the board need to approve such policies.
+  - Collaborate with the NixOS Foundation Board to set policies and priorities for high-visibility actions, events, and resources under the Foundation's care.
+    Both the SC and the board need to approve such policies.
   - Any delegation of responsibilities related to these areas also delegates communication with the Board on the relevant topics.
   - Be involved in the [board composition change process](#board-composition-changes)
 - Creation and Management of Teams:
@@ -47,7 +48,8 @@ It has the following responsibilities, which it may delegate as appropriate:
   - Ensure the existence and proper execution of processes for granting access to projects and resources.
   - As of the establishment of the SC, [official resources](https://github.com/NixOS/org/blob/main/doc/resources.md) and access details are listed in the repository.
 
-The SC has the authority to make decisions within the scope of its responsibilities; the restrictions on this authority are all explicitly listed in this constitution. The SC may make a decision to revoke delegation of a specific part of authority, if necessary — even if the authority in question has been with a certain team since before the current constitution.
+The SC has the authority to make decisions within the scope of its responsibilities; the restrictions on this authority are all explicitly listed in this constitution.
+The SC may make a decision to revoke delegation of a specific part of authority, if necessary — even if the authority in question has been with a certain team since before the current constitution.
 
 ### NixOS Foundation Board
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -27,3 +27,11 @@ For contributing see [here](./CONTRIBUTING.md).
...
   - [Nix Community Values](./doc/values.md)
 - [Calendar](./doc/calendar.md)
+
+## Drafting notes
+
+Text is written in Markdown.
+
+For the sake of human-readable diffs and patches, each line must contain at most one sentence.
+Long sentances may be broken into multiple lines at points of sub-sentance punctuation (e.g. comma, em-dash, semicolon) boundaries.
+Markdown explicitly ignores the location of line-breaks within paragraphs, so this convention does not affect the meaning of the text.
```
Ericson2314:
```suggestion
Markdown explicitly ignores the location of line-breaks within paragraphs, so this convention does not affect the meaning of the text.
(This rule a specific instance of the general principle of "semantic line breaks", discussed at greater length at <https://sembr.org/>.)
```

roberth: *CommonMark* behaves like this, but here we use GitHub Flavored Markdown, which
does
assign meaning to single line breaks.

So do we really want to adopt this rule?

The documentation team adopted this rule for CommonMark, in the manuals and nix.dev.

Ericson2314: @roberth I believe that for the last few years, GitHub has different rules for markdown files and GitHub comments. Comments are still
Line break
Sensative

But markdown files are not. Indeed, if one goes and toggles the rendered/pretty diff, it will say there are no changes.

roberth: Oh, TIL. Thanks

README.md
```diff
@@ -35,3 +35,4 @@ Text is written in Markdown.
 For the sake of human-readable diffs and patches, each line must contain at most one sentence.
 Long sentances may be broken into multiple lines at points of sub-sentance punctuation (e.g. comma, em-dash, semicolon) boundaries.
 Markdown explicitly ignores the location of line-breaks within paragraphs, so this convention does not affect the meaning of the text.
+(This rule a specific instance of the general principle of "semantic line breaks", discussed at greater length at <https://sembr.org/>.)
```
mdaniels5757:
```suggestion
(This rule is a specific instance of the general principle of "semantic line breaks", discussed at greater length at <https://sembr.org/>.)
```

mdaniels5757: Actually, I think I prefer:

```suggestion
(This rule is an instance of the general principle of "semantic line breaks", discussed at greater length at <https://sembr.org/>.)
```

Ericson2314:
An idea that just occurred to me is whether this information belongs more in `CONTRIBUTING.md` than `README.md`. What do you all think?

tomberek (approved):

cafkafk (changes requested):
I normally prefer to use line wrapping, but specially for things such as the constitution, I can't argue against the value of semantic line breaks, makes it more reviewable.

I would want to just get eyeballs from the rest of the SC on this, within reason (e.g., next few days), to make sure we're aligned on this, and that this didn't accidentally introduce any changes to the meaning of this document.

But in my reading, it didn't[1].

[1]: Also, while it is true that rendered markdown might in theory not affect the meaning of the text, this is less so the case as changes are introduced across lists and other elements.

doc/constitution.md
```diff
@@ -47,7 +48,8 @@ It has the following responsibilities, which it may delegate as appropriate:
   - Ensure the existence and proper execution of processes for granting access to projects and resources.
   - As of the establishment of the SC, [official resources](https://github.com/NixOS/org/blob/main/doc/resources.md) and access details are listed in the repository.
 
-The SC has the authority to make decisions within the scope of its responsibilities; the restrictions on this authority are all explicitly listed in this constitution. The SC may make a decision to revoke delegation of a specific part of authority, if necessary — even if the authority in question has been with a certain team since before the current constitution.
+The SC has the authority to make decisions within the scope of its responsibilities; the restrictions on this authority are all explicitly listed in this constitution.
+The SC may make a decision to revoke delegation of a specific part of authority, if necessary — even if the authority in question has been with a certain team since before the current constitution.
```
roberth, line 52: GitHub renders this *visually* as two paragraphs in the side by side view of the diff, while they are not.
This, in a way, reduces the reviewability.

roberth (approved):
I do not care for line breaks. Semantic line breaks should be called increased-entropy line breaks. Bike shed line breaks?

Clearly some people do care, so I won't block whatever this is.

doc/constitution.md
Ericson2314: You mean the thing where it makes a table with every line, and so more wrapping in one cell means vertical blank space in the other cell?

Yes it does do that, but it didn't really feel like "two paragraphs" to me, just line-based tabling.

JulienMalka (approved):
I'm okay with the spirit of this change, I have verified this is a no-op, I don't care too much about the details. 
I'd put the drafting notes in contributing.md, but again, I'm not going to bikeshed this.

Ericson2314:
> I'd put the drafting notes in contributing.md, but again, I'm not going to bikeshed this.

I already also came to the conclusion that that is better location than where I originally put it. Since @JulienMalka agrees, and no one else weighed in against this, I moved it there.

philiptaron (approved):
For sure, let's do it.

K900 (approved):
Y'all need Jesus.

cafkafk (approved):

Ericson2314:
We have unanimity!

infinisil (approved):
Nice

PR #199, opened 2025-11-08
Unify SC terms

KiaraGrouwstra:
As per our constitution:

> The Steering Committee (SC) [is] composed of 7 elected individuals to _represent_ the community

(emphasis mine)

We seek to achieve this using proportional representation, which (as per Wikipedia):

> is achieved by any electoral system under which subgroups of an electorate are reflected proportionately in the elected body.

Currently, this proportionality is distorted by our distinct term durations induced by our staggering mechanism, by:

1. distinct term durations (for full elections or early terminations) granting some voter contingencies representation for 2 years, others for 1 year, despite ranked voting already offering secondary preferences as a mechanism to deal with superfluous votes
1. limiting the granularity of the reflected voter preferences, as not necessarily all seats would be available in the annual election

This proposal resolves this by removing the stagger effect, unifying the terms, reasoning that the electorate would nevertheless remain empowered to reelect the incumbent candidates.

Knowledge transfer, one potential intent behind the term stagger, I imagine could be performed without retention of formal titles, if the electorate were to pick different representatives.

Note that in unifying the term duration, I had to drop one of the currently used term durations.
In this case I opted to retain the 1-year term duration, such as to best reflect community preferences without requiring an increase in the number of elections, although this preference is secondary to the proposal (and could be decoupled).

Further note that I left the maximum number of consecutive terms intact at 2, which under the chosen term duration shortens the maximum consecutive duration one can be elected for.

edit: to clarify, this PR got marked closed as an unintended consequence of an attempt to clear out duplicate notification references from draft commits for #200 - we can refile as needed.

### DIFF
--- a/doc/constitution.md
+++ b/doc/constitution.md
@@ -106,8 +106,7 @@ An exception to that is when there is when the [conflict of interest balance](#c
 ### Composition and Appointment
 
 The SC consists of 7 elected members.
-Regular elections are once a year, with half (alternating rounding every year) of the seats up for election,
-such that each member holds a seat for one term of at most 2 years before it is up for election again.
+Regular elections are once a year, with each member holding a seat for one term of 1 year before it is up for election again.
 Members can at most serve two consecutive terms. There is no lifetime maximum.
 
 The previous SC appoints an Election Committee (EC) of at least 3 people to administer the election. If the SC cannot come to supermajority agreement on the list of EC members, they will elect an EC using the same tallying system as the previous SC election, where only SC members can vote.
@@ -161,7 +160,6 @@ In any election where seats with different end-of-term dates are available, winn
 
 #### Special considerations for the first SC election
 
-- To stagger the terms, the first election will appoint half the members rounded up with 2-year terms and the remaining members with 1-year terms.
 - The EC for the first election will be the members of the [Nix Constitutional Assembly (NCA)](https://github.com/nixos/nix-constitutional-assembly?tab=readme-ov-file#members), which are thus ineligible to be elected for the first SC
 - All nominees become candidates for the first election, because there's no existing SC that would decide over disqualifying candidates.
 - And to help the first elected SC establish itself, members of the NCA are integrated into the SC as non-voting members for 6 months after the appointment. The NCA ceases to exist at the moment of SC appointment.
@@ -174,11 +172,11 @@ Furthermore, there is a hard maximum of two members with the same CoI at any tim
 
 #### Special Elections
 
-In the event of a resignation or other loss of a steering committee member (including a [removal for conduct](#removal-for-conduct)), a special election for that position may be held if the SC deems it necessary, or the SC does not have half of the normal size. Otherwise, the missing positions are assumed to be abstaining from all the SC votes. A committee member elected in a special election will serve out the remainder of the term of the person they are replacing, regardless of the length of that remainder.
+In the event of a resignation or other loss of a steering committee member (including a [removal for conduct](#removal-for-conduct)), a special election for that position may be held if the SC deems it necessary, or the SC does not have half of the normal size. Otherwise, the missing positions are assumed to be abstaining from all the SC votes. A committee member elected in a special election will serve out the remainder of the term of the person they are replacing.
 
 #### Full Reelections
 
-A simple majority within the SC may call a reelection of the entire SC based on perceived loss of confidence. In this case, it also has to be decided whether this election is considered a special election for the remainders of all the corresponding terms, or an initial election for full 2-year terms for half of the seats rounded up and 1-year half-terms for the remaining seats.
+A simple majority within the SC may call a reelection of the entire SC based on perceived loss of confidence. In this case, it also has to be decided whether this election is considered a special election for the remainders of all the corresponding terms, or an initial election for full 1-year terms.
 
 ### Removal for conduct
 


K900:
Another point that was raised re: this is [here](https://discourse.nixos.org/t/proportionality-and-the-staggered-terms-in-sc/70563), which basically means that any attempt to have both proportional voting and staggered terms will result in every other SC being actually disproportional. I support the proposal and will happily hand over my seat next year if it passes.

doc/constitution.md
```diff
@@ -107,7 +107,7 @@ An exception to that is when there is when the [conflict of interest balance](#c
 
 The SC consists of 7 elected members.
 Regular elections are once a year, with half (alternating rounding every year) of the seats up for election,
-such that each member holds a seat for one term of at most 2 years before it is up for election again.
+such that each member holds a seat for one term of at most 1 year before it is up for election again.
```
K900: Can drop the "at most" here, I guess?

JulienMalka:
I would be in favor of such a change. We maybe have to also explicitely include provisions for the transition.

KiaraGrouwstra:
@JulienMalka: transition considerations may depend a bit based on whether a term of 1 or 2 years is settled on. I imagine the incumbent SC could figure out the transition.

Ericson2314 (changes requested):
I strongly oppose getting rid of the staggering until we we are still trying to bootstrap the existence of a steering committee at all. (And to be clear, I am now a 1-termer up for reelection, so if anything, I personally would "benefit" from this.) I think building a culture of *having governance at all* (dovetailing with the task of trying to create a vision that is both sufficiently ambitious and uniting) is the most important thing. And staggering for the strongest continuity as governance struggles to be born I believe immensely helps with both of those.

Only once we have governance and a vision "up and running' should we even consider getting rid of staggering.

------

Side note:

In generally I think too many people are trying to tackle our problems through constitutional gimmicks. At the nation scale, I think that is actually very appropriate --- e.g. America's poorly-designed elections I believe are a severely underrated source of its political problems. But the Nix community is orders of magnitude smaller, and as much as people want to project their experience of real-world politics onto Nix politics, our problems are not, in fact, the same.

To make my point about scale a bit more tangible: "Marriage counseling" is closer in scale to us than "continent-wide superstate (US, EU) politics", "large western European country politics" or even "the Netherland's politics". And indeed I do think more emotions- and empathy-first, or therapeutic, approaches are a more appropriate method too.

K900:
In that case, would you support switching elections to approval voting? That would lose some of the proportionality, but will at least avoid skewing the resulting SCs quite as strongly.

Ericson2314:
I agree that ranking everyone in total order is annoying, but I also agree with @7c6f434c that proportionality is of *utmost* importance. The good news is that there are other things options like https://en.wikipedia.org/wiki/Proportional_approval_voting that exist.

That doesn't address your point of staggering skewing the results, but I think that is *already* solved by the factions and staggering not correlating. For example:

- The original 2 year people from 1st year were Gabby, Jan, Robert, and I, which I think is fairly ideologically balanced. The outcome that made it seem imbalanced was because of Gabby and Jan resigning, *not* because of anything the constitution did.
- Going to the 3rd year, you, Christina and Tom are the ones automatically staying on, which also strikes me as ideologically balanced.

So 2/2 times, I think the election results did not unduly bias any one faction. We empirically are having our cake (continuity from staggering) and eating it (no crazy ideologically oscillation from mandatory staggering) too. Problem avoided!

KiaraGrouwstra:
uneven terms do cause oscillations, which arguably not only distort proportionality but further facilitate destabilizing (see the thread @K900 linked).

> the factions and staggering not correlating

> no crazy ideologically oscillation from mandatory staggering

the aforementioned limited granularity of voter preferences supported by the staggering, would structurally disadvantage e.g. smaller constituents accounting for perhaps 1/7 yet not for as much as 1/4~1/3 of the electorate (or whatever number of seats is up in any given staggered election now).

considering the number of topics up for discussion in a project as vast as nix's, i think there could be constituents on many axes we might not even conceive of yet now.

heck, given n binary axes, we would already end up with 2^n potential constituents, so the existence of smaller constituents would seem not that far-fetched.

> I think building a culture of _having governance at all_ (dovetailing with the task of trying to create a vision that is both sufficiently ambitious and uniting) is the most important thing.

i would personally consider the pluriformity reflected by proportional representation a feature, but if we had to create a vision, i would argue the value of such a vision derives from it reflecting our community - which proportionality helps achieve.

i feel like i may be missing @Ericson2314's point still tho.
what continuity would require titles, if knowledge transfer could be handled, newly elected delegates could similarly contribute to a vision, and we so far all regard proportionality as positive?

tomberek:
In the spirit of "[Stable evolution over stagnation or chaos](https://nixos.org/values/#community-values-stability)", staggered concept encourages people to stay around longer, to be committed for a longer term. It also allows for evolution, rather than a fresh set of people who have to constantly figure out how to work together. The aggregate average of the community's vision doesn't become the aggregate average of an elected SC; people and beliefs and visions are not fungible or transferable or averagable in that way - we are not robots or high-dimensional vector representations. In fact, it is the people that we need to focus on and long-term leaders we should be creating "[People come first ... Individuals gain trust and status by doing the work.](https://nixos.org/values/#community-values-people)". A stagger allows people to work with others in a leadership capacity, and hand-off the vision to a new generation over time.

Also, be wary of unintended consequences. Consider that it is very common in governance and politics for things to backfire and have the opposite outcome than what you may expect. Part of the reason of proportional voting was to protect minorities, to ensure they would have representation rather than be lost in the "average". Part of the reason to have staggering is to prevent wild swings of sentiment, it is a low-pass filter. And like any low-pass filter, it introduces a delay in the response, for the benefit of preventing fast changes that introduce chaos and fragility. Similarly, the high requirement of super-majority to change the constitution is there to smooth out changes to the rules; fast-changing rules lead to chaos. We're not playing Fluxx/Nomic, we are trying to [develop, propagate, and promote the adoption of a purely functional software deployment model](https://github.com/NixOS/foundation/blob/master/by-laws/Oprichtingsakte_NixOsFoundation.en.pdf)

I am currently opposed to this constitutional change. I'll also note that I was opposed to removing the stagger even before my re-election.

piegamesde:
Every governance structure design decision is a tradeoff between costs and benefits. I'd argue, that in our case, the staggered voting has higher costs and lower benefits than is worth.

**High costs:** The primary cost is that unless we perfectly follow the initially set out cadence, proportionality is reduced. Practical experience has shown us that being on the SC is a tough job and that many people won't be able to hold onto their two year terms, yet alone make it through the full year. Prior experience with other high-intensity teams like moderation shows that there is no reason to expect this to change in the future. But moreover, due to systemic discrimination in our societies applying a higher baseline pressure, people from marginalized backgrounds are at higher risk at burning out. This means that in its current form, **the staggered voting inherits structural discrimination in form of a bias against minorities**

This is a major issue, and if we are to keep staggered voting then we *must* find alternative solutions which preserve proportionality.

**Low benefit:** The primary reason for the staggered voting is to provide stability in governance, allowing a smooth handover and knowledge transfer. I'd argue that thanks to the current voting system, there is always a good chance that several current SC members being reelected should they run. And even if all people change, the general direction of the voting result is unlikely to change in a major way. About the knowledge transfer, there are many alternative ways to achieve that which don't come with the aforementioned downsides.

------

I used to be in favor of the staggered voting, but now that we tried it out for a year and having seen how it works out in practice, I can't justify supporting that position anymore. The only remaining argument in favor of keeping this current status quo would be inertia and structural stability, but given how easy this change should be to implement I don't think this is justified.

Ericson2314:
> I used to be in favor of the staggered voting, but now that we tried it out for a year and having seen how it works out in practice, I can't justify supporting that position anymore.

This is logically incoherent. We didn't try our staggering for a year. We just staggered for the first time 2 weeks ago. The knowledge transfer benefits don't apply to SC 1, they apply to SC 2. We won't know how much benefit increased continuity gives until we have more experience with the new SC.

piegamesde:
Thank you for completely ignoring my point. My point is that the downsides *already* outweigh all hypothetical benefits, *and* that we could get those benefits differently without said downsides.

7c6f434c:
Arguably you could make a wider point: providing a natural resignation point mid-mandate is not even viewpoint-neutral (depending on the kind of the goals one sets, burn-out risk also differs, which leaks into proportionality on the kind of goals _everyone_ agrees SC has).

On the other hand, I would estimate that there was context loss on the handover NCA → SC in a way that will probably repeat itself with full rotation, so we should wait a few months to see how it goes with staggered handover. Ending up with polarisation as a foundation of institutional memory is not free of moral hazards — and we need to understand whether anything worked in staggered transition to design an alternative solution.

7c6f434c:
I guess enough time after the election has passed and I do not have propriety of the office to maintain w.r.t. taking sides anymore.

> In the spirit of "[Stable evolution over stagnation or chaos](https://nixos.org/values/#community-values-stability)", staggered concept encourages people to stay around longer, to be committed for a longer term. It also allows for evolution, rather than a fresh set of people who have to constantly figure out how to work together.

@tomberek if we look at the elaboration, which in case of NCV has no less weight than the header, I do not believe that there is a well-defined interface that can be stabilised that your argument is protecting. 

And having a group of people where _half_ of them have already learned to work together has its own special costs in the sense of asymmetry but not enough of a gelled core to provide an «assimilation target».

I think inertia vs nonlinear burnout risks is a trade-off where neither side should be blindly discarded, and I agree that  institutional knowledge should be considered in the design — but your reference to the value here seems headline-oriented to me.

PR #200, opened 2025-11-09
retire value stable evolution

KiaraGrouwstra:
One of the values we inherited from the interim Nix constitutional assembly is '[stable evolution over stagnation or chaos](https://nixos.org/values/#community-values-stability)'.

This value has since been used to argue against:

- [accountability](https://discourse.nixos.org/t/sc-meeting-2025-10-08-17-00-utc/70609) (why we even wanted elected representatives)
- [proportional representation](https://github.com/NixOS/org/pull/199#issuecomment-3506958459) (characterized as [of utmost importance](https://github.com/NixOS/org/pull/199#issuecomment-3506788693))
- (as per above) [protecting from destabilizing take-overs](https://discourse.nixos.org/t/proportionality-and-the-staggered-terms-in-sc/70563) (how was the value even intended?)

To be clear, if on a technical decision we used stability as a tiebreaker — then I get it.

When the community indicated the board last year seemed out of touch, the board opted to listen and step aside to make room for elected representatives.
If this value, however its intent, ends up functioning as it has though, I call that a regression, and propose reverting it.

### DIFF
--- a/doc/values.md
+++ b/doc/values.md
@@ -59,13 +59,3 @@ We build trust primarily by working together on Nix projects.
 We are a global community, and disseminating information and maintaining processes can be difficult. We are also a large project with a lot of hard and repetitive work. Therefore, we value automation over toil, while recognizing that not all toil can be automated. Automation reduces toil, but people are still accountable. Adding new toil needs a very strong justification.
 
 We build automation and processes that make the best use of our contributors' limited time and energy.
-
-### Stable evolution over stagnation or chaos
-
-Openness to new ideas and evolution is part of what made Nix great. We continue to foster that evolution while encouraging development of re-usable building blocks and well-defined, stable interfaces.
-
-We value experimenting with designs and concepts, and folding successful experiments back into continuous improvement for stable components.
-
-The larger the impact an action has, the more care and discussion is warranted before taking the action.
-
-Our leaders have a duty to find, support, and promote new contributors — and eventually step aside for new leaders.


piegamesde:
I do think the value is reasonable as it is written, and as I argue over at https://github.com/NixOS/org/pull/199#issuecomment-3508319662 I don't see it in conflict with one-year term voting.

We currently have people with a serious track record of avoiding accountability in our governance (including but not limited to the SC), and knowing them I doubt that changing some words here or there will help against that. @KiaraGrouwstra, I kindly ask you to reopen the other PR and optionally also close this one, as the current path is unlikely to be successful in its actual goal.

cafkafk (commented):
I wrote my take on this here https://discourse.nixos.org/t/proposal-stagnation-and-chaos/71919/2.

The moral of the story being that removing this might make sense, but what I'd really prefer is to fix that problem that these lines were written to fix in the first place. 

Saying you're open is cheap, writing the actual thing that commits you to being open is much harder! Saying to make space is not the same as making actual space!

And the idea that the more impact an idea has, the harder it should be to make it is essentially a line that cements total stagnation. Why would we want the most important changes to be the hardest to make? Surely our goal should be the opposite, to make it as easy as possible to do as much as possible with as little as possible. That is of course, unless our goals were to never change (which is suspect is the underlying motivations behind those lines, a kind of “let's stay stuck in the cycle of community leadership crisis and complaining without solving forever.”).

That's why I'm not moving to block this as well before it's in a perfect shape that suits me. But that said, I would prefer if we went for the absolute worst lines first:

> The larger the impact an action has, the more care and discussion is warranted before taking the action.

These lines are irredeemable, the intent is to minimize harm, but the reality is that it maximizes the entrenched harm. Better yet if we rewrote this as something more useful.

> The more impact an action has, the more important it is to triage whether it's obviously harmful, and if not, to collectively support the expedient resolution of said action into a valuable outcome.

This is just a very bad first approximation, but it feels like the real solution isn't to say “this is really hard and scary, so I'm going to make it exceptionally hard for you to do the right thing” but instead it should be “this is extremely hard and scary, let's do our best to make sure we can confirm doing this is a good thing, and if that's the case lets all collectively do everything in our power to carry that burden of hard and scary together”.

roberth (changes requested):
I oppose a complete removal. We could perhaps revisit individual changes later.

wolfgangwalther:
Changing any of the values substantially requires:

> Substantial amendments to the Nix Community Values require 90% agreement in a poll among eligible voters.

This is nothing that the SC can even vote on. It is *very* unlikely to happen - you can assume the values to be set in stone forever.

(of course, the SC could *change the constitution to allow them to change the values* - but you'd have to propose that change first)

Gabriella439:
> (of course, the SC could _change the constitution to allow them to change the values_ - but you'd have to propose that change first)

https://github.com/NixOS/org/pull/202

KiaraGrouwstra:
@cafkafk's amendment of the value sounds like a step forward.
Even using such a rephrasing tho, I could see still see some room for someone in a position of power to unilaterally judge that a dissatisfied community, in their act of discontent, must not have applied sufficient care in their considerations to warrant change.

Some responses wondered if this PR may have targeted symptoms, rather than an actual issue. To clarify, I [do](https://github.com/NixOS/SC-election-2025/issues/482) in fact support responsive and accountable governance, and believe considering incumbency as inherently trumping such considerations to be unproductive. I would rather see considerations be taken at their merit, so we could actually evolve, and see enshrining incumbency as running counter to this.

tomberek:
The values are supposed to make you reconsider; that is why they exist. They make you think and self-reflect using a set of principles. If we found them to never be referenced or used, they would be meaningless. They need interpretation and are not solid rules - they help guide conversations and help weigh uncertain options.

In this case, the value reminds us about how we manage change. We allow for change, but not upheaval. We limit frantic alterations, but allow for evolution.

At the moment, we suffer from too much change, not from too little. A new governance paradigm every year, brand new Board each year (except for Ron, thanks!), new SCs, new teams being created .... and now we have several proposals to keep changing even more about the constitution, changed voting mechansims, proposals for a full reset of SC, Value changes. Frankly, this seems to be a bunch of rules-lawyering that should go through a low-pass filter ... and we have one, the SC.

This provides stability, something our organization sorely needs. If people, orgs, companies, and communities want to invest in Nix/Nixpkgs/NixOS, we need to have a track record of enough stability for people to feel their investment won't have the rug pulled out from under them each year.

Even if you disagree about the current situation and claim we have stagnated, then at least the opposing viewpoints should be able to agree to aim for some middle ground - the Value as written.

Summary: I oppose this change.

KiaraGrouwstra:
> upheaval
> frantic alterations
> stability
> rug pulled

I would consider these loaded language.
What is it that makes you feel new ideas would be inferior to existing ones?

Our codebases for one see their fair share of changes. I believe we tend to generally consider those improvements (tho on occasion we might fix a regression).
Why take the opposite stance in governance?

Listening to the NCA's presentation, I got the sense this value stemmed mostly from the flakes situation, with its introduction focusing on experimentation then stabilizing. @tomberek's reading of the value, by contrast, seems to clash with that idea of experimenting. My best attempt to reconcile those is the NCA did not focus as much on governance questions where the incumbent option might be (to various extents, depending on thresholds picked) worse.

doc/values.md
```diff
@@ -59,13 +59,3 @@ We build trust primarily by working together on Nix projects.
...
-
-### Stable evolution over stagnation or chaos
-
-Openness to new ideas and evolution is part of what made Nix great. We continue to foster that evolution while encouraging development of re-usable building blocks and well-defined, stable interfaces.
-
-We value experimenting with designs and concepts, and folding successful experiments back into continuous improvement for stable components.
-
-The larger the impact an action has, the more care and discussion is warranted before taking the action.
-
-Our leaders have a duty to find, support, and promote new contributors — and eventually step aside for new leaders.
```
wolfgangwalther, line 71: I'm not a friend of removing that last sentence - I think it is possibly more important than ever before. We have it written down, but neither of the two aspects have been happening, so far. We should focus on interpreting *all* of what's written in that value and not just pick the header whenever we want to argue against something we define as "chaos".

Gabriella439:
> At the moment, we suffer from too much change, not from too little

Do we?  What has *actually* changed over the last year?  All of the Constitutional proposals so far have been shot down, even the less controversial ones.

I'm not necessarily disagreeing with the other points you made, but the reason I'm highlighting this one specific point is that I don't want us to conflate *proposing changes* with *actual changes* because if "proposing changes = instability" then the stability value can be weaponized to shut down even the *debate and/or discussion* of possible changes.

wolfgangwalther:
> What has _actually_ changed over the last year?

We are now without moderation team. I consider this a pretty hefty change.

jfly:
+1 to what @wolfgangwalther said. It might help if folks stop using the word "change" unqualified here, it feels like people are talking past each other. Taking some examples out of context (sorry):

> At the moment, we suffer from too much change, not from too little

I believe that when @tomberek uses the word "change" here, he's referring to "unstable change" or perhaps "undesirable change".

> What has actually changed over the last year?

Here I believe that @Gabriella439 means something like "what has actually changed for the better?".

I personally want stable evolution ("good change") such as retiring or stabilizing experimental features (the new cli and flakes come to mind).

On the other hand, I don't desire to be in a community with unstable ("bad") change, such as seeing nearly an entire team resign in protest of the SC's actions (not blaming anyone here, just citing it as an example of unstable change).

The rub is that when folks are frustrated with the status quo or feel unsafe, they don't always care about stable evolution: when my house is on fire, I don't care if someone breaks a window to save someone inside.

PR #202, opened 2025-11-10
Change amendments to the values to require an SC supermajority

Gabriella439:
There are two main rationales behind this change:

- It doesn't make sense for amendments to the values to require a higher threshold than amendments to the Constitution, given that the Constitution governs amendments to the values

  Whether or not people believe that this Constitutional amendment *should* pass, the fact that this amendment *could* pass with a 5/7 vote highlights why the 90% requirement does not actually exist in practice.

- The first draft of the values was not approved by 90% of the community

  It seems weird to privilege the first draft of the values when it never had to clear the same bar as amendments to the values.

### DIFF
--- a/doc/constitution.md
+++ b/doc/constitution.md
@@ -95,7 +95,7 @@ Furthermore, the SC decides on the Election Committee (EC) with a 2/3 supermajor
 
 Disqualifications of candidates in an election requires supermajority among the currently serving SC members.
 
-Substantial amendments to the Nix Community Values require 90% agreement in a poll among eligible voters. Deciding that an amendment is not substantial can be done by unanimity among a full SC.
+Substantial amendments to the Nix Community Values requires a supermajority of a full SC.
 
 #### Ordinary decisions
 


tomfitzhenry:
The mandate for the Values document, decided by the electorate, is a feature that ensures these foundational principles reflect a broad consensus.

The fact that its amendment bar is higher than the Constitution's can be viewed not as an inconsistency, but as an intentional elevation of the level of agreement required for our most basic principles.

We are entrusting the Steering Committee to uphold this spirit, and a stronger alignment in our governance might eventually see a similar high mandate considered for the Constitution itself.

tomfitzhenry:
> The first draft of the values was not approved by 90% of the community
>
> It seems weird to privilege the first draft of the values when it never had to clear the same bar as amendments to the values.

I agree, other PRs suggest there might be some clauses which don't have 90% support, but would need 90% to remove. In other words, there might be clauses/sections which would only need 11% support to remain.

I think it's worth revisiting the Values, but still via direct electorate approval (in some form) rather than via SC.

KiaraGrouwstra:
This sounds to me like a step in the right direction.

Engraving things in stone (as the never-passed 90% threshold implies) would significantly hamper evolution of the document, particularly at a point when it had not been tested in practice, effectively overruling democratically elected bodies by a body that was not elected itself.

I would propose things be set in stone only after we grow confident over the years these are indeed what works best for our community.

I set out my further considerations on this at https://github.com/NixOS/org/pull/203#issuecomment-3516170643.

7c6f434c:
> effectively overruling

Note that structurally, NCV can force consideration of some facet of an issue, but cannot force a final decision if SC is firm. And pretty much always there is a trade-off between values involved anyway, so the values are really not suitable to overrule the final decision.

PR #203, opened 2025-11-10
Make the NCV amendment requirement self-referential

rhendric:
If for whatever reason the requirement difference between ‘doing the thing’ and ‘changing the rules that restrict doing the thing’ is seen as an inconsistency or weakness to be exploited, this is the natural way to close the loophole while preserving the original intent.

This is, of course, an alternative to #202, though I'm not sure either amendment is really needed.

### DIFF
--- a/doc/constitution.md
+++ b/doc/constitution.md
@@ -95,7 +95,9 @@ Furthermore, the SC decides on the Election Committee (EC) with a 2/3 supermajor
 
 Disqualifications of candidates in an election requires supermajority among the currently serving SC members.
 
-Substantial amendments to the Nix Community Values require 90% agreement in a poll among eligible voters. Deciding that an amendment is not substantial can be done by unanimity among a full SC.
+Substantial amendments to the Nix Community Values require 90% agreement in a poll among eligible voters.
+Substantial amendments to this paragraph of the constitution carry the same requirement, in addition to any other constitutional requirements for making that change.
+Deciding that an amendment of either document is not substantial can be done by unanimity among a full SC.
 
 #### Ordinary decisions
 


Gabriella439:
I'll bring up the same point I brought up in #202, which is that it's weird for amendments to the values to be held to a higher bar than the initial draft of the values (which was not approved by 90% of the voters).  It implies an imprimatur on the current values which never actually existed.

rhendric:
> It implies an imprimatur on the current values which never actually existed.

I can see where you're coming from, but I don't think it implies this so much as it establishes that the important thing is that the values aren't a rapidly moving target. The initial expression of the values served as a signal to people who might be considering leaving or forking the community — either you can see yourself working within some radius of these values, or you prefer to try your luck doing something else. They didn't have to meet a 90% bar because most of the benefit was in planting a flag, not determining the exact center of mass of the community based on some inevitably controversial weighting and selection process.

Now that the flag is in place, and people have made their choices with respect to staying or leaving, I think moving the flag around from under people should be much rarer and have to meet a higher threshold for accuracy than the initial flag planting did, because motion is far likelier to fling currently happy people away than it is to bring previously unhappy people back — the latter don't have a reason to be paying attention anymore, and even if they are, there's a higher trust gradient to climb now.

KiaraGrouwstra:
This would substantially bias things toward ideas just because they're old - despite learning processes empowering more _recent_ insights to be better informed.
As such, weighing ideas by their actual perceived merit seems significantly more sensible.

As such, if the incumbent values turn out flawed in unintended ways at some point, engraving them (as this proposal would) might serve mostly to divide the community into forks.

That raises the question: is it more valuable for the Nix community to stablize a document the NCA settled on in 2024, or to be able to gradually figure out how we can make things work for us as a community?

Listening to the NCA presenting the values, it sounds like the point was for the values to have majority support, and actually be shared by the community.

The current proposal could make it such that, if the community sooner or later finds out the document does not properly reflect how they actually feel, a >10% minority could maintain the (quite possibly sub-optimal) status quo, rather than the (if unintentionally) >33% minority required now.

Anecdotally, that does not seem much of a stretch, as #200 demonstrates that the current wording of the values within the SC at least already does not reach the criterion of being deemed optimal by 90% of its members.

Should we really want that?

At NixCon 2024, the NCA explained their intent:

> Our ultimate goal was to was to get people with very different opinions, perspectives, outcomes, interests, to be able to work together within this community.

In my opinion, that is not achieved by having 89% of the community ruled by a dated piece of paper that, despite its best intentions, turns out could be improved a bit further still.

Ericson2314:
Thank you @rhendric I like this solution. I've consistently been impressed with the work that the NCA did, and the system they've set up for us. I like the values too, and they strike me as things that are reasonable and uncontroversial.

I do agree this fixes the bug and reflects the NCA's intent. (And I think the "bug" framing is appropriate because there isn't some other immenant decision that hinges on the bug in an ironclad way.)

Ericson2314 (approved):

JulienMalka:
I agree with @Gabriella439 that there is no reasons the values would require 90% of the electorate to be changed when they were not approved by 90% of the electorate to begin with. This amendement would only make sense if we first subject the initial set of values to a referendum.

Ericson2314:
Put another way, I would rather hold a referendum then than have a weaker values document. Independent of how we got here, I like having a stronger values document (like we do now) going forward.

tomberek:
This does add some clarity and is the intent of the document (strongly held).

But I'm not sure if it is necessary (weakly held).

roberth (approved):
I agree with @rhendric's explanation in terms of flag planting, and I'd like to highlight:

> They didn't have to meet a 90% bar because most of the benefit was in planting a flag, not determining the exact center of mass of the community based on some inevitably controversial weighting and selection process.

Furthermore, the values aren't law.

> Values are not a Code of Conduct, they do not define governance structures, and they do not provide specific policies. Instead, values inform the decisions we make about these things.

&mdash; [Values - What is a value?](https://nixos.org/values/#community-values-whatis)

And even then they have room for interpretation.

That said, I do believe an 90% threshold is high for a vote that will inevitably politicized by correlation with other things/people. Nonetheless, a change to the values should carry broad agreement.

Gabriella439:
> That said, I do believe an 90% threshold is high for a vote that will inevitably politicized by correlation with other things/people. Nonetheless, a change to the values should carry broad agreement.

Well, if this change were approved you would no longer be able to adjust the threshold except by 90% agreement of the voters

roberth:
I wouldn't be saying that if I didn't want to signal that it's possible to discuss that, wouldn't I?

PR #205, opened 2025-11-12
xrefcheck: Only check local links

infinisil:
We get frequent failures due to server problems when checking external links


infinisil:
Case in point..

image

winterqt (approved):
thank god

PR #207, opened 2025-11-17
Create commentary (including flowchart) about constitution

Ericson2314:
As the complex flowchart hopefully makes clear, I think the constitution is more complicated than it might first appear, especially after recent changes.

This commentary is supposed to make all this explicit.

(Reviewer note: use rendered diff to view mermaid diagram.)

### DIFF
--- a/doc/constitution-commentary.md
+++ b/doc/constitution-commentary.md
@@ -1,0 +1,94 @@
+# Constitution Commentary
+
+This document contains a non-normative, non-binding commentary on the constitution, explaining some subtleties.
+
+## SC Election and Removal procedure
+
+The state machine of the portions of the constitution that govern the composition of the SC (what seats are filled, who is in which seat) is especially complex.
+The following flow chart diagrams this state machine:
+
+```mermaid
+flowchart TD
+    Start([Regular SC Governance])
+    End(["Regular SC Governance (with new members)"])
+
+    %% Note that order effects layout, at least informally.
+    %% Current layout chosen in part because graph renders planar (no edges cross).
+
+    %% Regular elections
+    RegularElec[Regular Election]
+
+    Start -->|some terms end| HalfVacation[[3-4 seats to be vacated]] --> RegularElec
+    RegularElec -->|seats filled for fresh 2-year terms| End
+
+    %% Special Elections
+    SpecialElec[Special Election]
+
+    Start -.->|optional, if <7 filled| SpecialElec
+    SpecialElec -->|winners serve remainder of terms| End
+
+    %% Single seat vacation
+    Removal[Vote for removal for conduct]
+    OneMoreVacation[[one additional seat vacated immediately]]
+    Required[Vote for Special Election or dissolution]
+    AllVacated0[[all seats to be vacated]]
+
+    Start -.->|optional| Removal
+    Start -.->|resignation or other loss| OneMoreVacation
+    Removal -->|succeeds by supermajority| OneMoreVacation
+    OneMoreVacation --> VacancyCheck
+    Removal -->|fails| Start
+    VacancyCheck{Vacancy count?}
+    VacancyCheck -->|≥4 filled| Start
+    VacancyCheck -->|<4 filled| Required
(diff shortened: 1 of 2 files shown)


KiaraGrouwstra:
Thanks, this felt kind of insightful.

I wondered if the states on half the seats being filled could be combined somehow.

tomberek:
@7c6f434c for review

JulienMalka:
Right, the charts looks correct to me. @wolfgangwalther you have usually a correct interpretation of constitution, wdyt?

As for this being the right place for this kind of content, I am not sure yet; I'll sleep on it and no better option emerges I'll approve.

Ericson2314:
> As for this being the right place for this kind of content, I am not sure yet; I'll sleep on it and no better option emerges I'll approve.

I would like it to be in this repo at least, so future revisions to the constitution can modify the text and flow-chart in one step. But other than that, I have no preference.

philiptaron (approved):
This chart matches my understanding of what the constitution says regarding elections.

wolfgangwalther:
> @wolfgangwalther you have usually a correct interpretation of constitution, wdyt?

I'll defer to @7c6f434c, who certainly has a much better idea than I do.

wolfgangwalther:
Looking at this I can't help but think "this is too complicated - we need a simpler constitution"...

One thing that should probably be done: When the number of SC members falls below 4, no-confidence should be automatically triggered. Reduces the number of choices and also resolves https://github.com/NixOS/org/pull/208#discussion_r2535490621.

Ericson2314:
@wolfgangwalther I agree it is too complicated.

But I think it is nice to merge the diagram before, so when we then make the simplification PR, we upgrade the diagram too, and the reviewers of the next PR can review the changes to the diagram and the changes to the constitution at the same time to make sure they match.

Conversely, if we make the diagram after we make the simplifications, we run the risk of voting on text whose meaning is not yet clear...because we don't have the diagram yet!

(I certain learned a lot about the constitution in making this diagram --- I simplify would not have been able to get it all straight without the diagram.)

doc/constitution-commentary.md
```diff
@@ -0,0 +1,76 @@
...
+    Start -.->|optional| Removal
+    Start -.->|resignation or other loss| OneMoreVacation
+    Removal -->|succeeds by supermajority| OneMoreVacation
+    OneMoreVacation --> VacancyCheck
+    Removal -->|fails| Start
+    VacancyCheck{Vacancy count?}
+    VacancyCheck -->|≥4 filled| Start
+    VacancyCheck -->|<4 filled| Required
+    Required -->|all members propose special election| SpecialElec
+    Required -->|any member proposition dissolution| AllVacated0
```
mdaniels5757:
```suggestion
    Required -->|any member proposes dissolution| AllVacated0
```

doc/constitution-commentary.md
```diff
@@ -0,0 +1,76 @@
...
+    AllVacated0 -->|special election wins via vacant seat votes| SpecialElec
+
+    %% Dissolution
+    Dissolution[Vote to dissolve the SC<br/>vacant seats vote yes]
```
mdaniels5757: If the syntax allows it, it's be nice if "vacant seats vote yes" were enclosed in ().

doc/constitution-commentary.md
```diff
@@ -0,0 +1,76 @@
...
+    AllVacated1[[all seats to be vacated]]
+    ElectionType[Vote between election types<br/>vacant seats vote special]
```
mdaniels5757: If the syntax allows it, it's be nice if "vacant seats vote special" were enclosed in ().

doc/constitution-commentary.md
Ericson2314:
```suggestion
    Dissolution["Vote to dissolve the SC<br/>(vacant seats vote yes)"]
```

doc/constitution-commentary.md
Ericson2314:
```suggestion
    ElectionType["Vote between election types<br/>(vacant seats vote special)"]
```

PR #208, opened 2025-11-17
Slight wording tweaks in constitution

Ericson2314:
**This should be a pure refactor. The meaning of the constitution should not be change. Any intentional meaning changes should be perused separately in other PRs.**

Please see each commit for details of the rewording in question.

These were made at the same time as #207, based upon my struggling to understand the meaning of the constitution. I also have some other pure-refactor changes on top of these, but those are still a work in progress, so I will PR those separately if these pass.

### DIFF
--- a/doc/constitution.md
+++ b/doc/constitution.md
@@ -200,9 +200,13 @@ If it is impossible to find sufficient members to resign, all members with that 
 
 #### Special Elections
 
-In the event of a resignation or other loss of a steering committee member (including a [removal for conduct](#removal-for-conduct)), a special election for that seat may be held if the SC deems it necessary, or the SC does not have half of the normal size.
+In the event of resignations or other loss of steering committee members (including [removals for conduct](#removal-for-conduct)), the SC would have less than all 7 seats filled.
+
+When one or more seats are vacant, a special election may be held for those vacant seats.
 Otherwise, the missing seats are assumed to be abstaining from all the SC votes.
 A committee member elected in a special election will serve out the remainder of the term of the person they are replacing, regardless of the length of that remainder.
+
+A special election for vacant seats may be held if the SC deems it necessary, or the SC does not have half of the normal size.
 
 #### Full Reelections
 
--- a/doc/constitution.md
+++ b/doc/constitution.md
@@ -197,8 +197,8 @@ If it is impossible to find sufficient members to resign, all members with that 
 
 #### Special Elections
 
-In the event of a resignation or other loss of a steering committee member (including a [removal for conduct](#removal-for-conduct)), a special election for that position may be held if the SC deems it necessary, or the SC does not have half of the normal size.
-Otherwise, the missing positions are assumed to be abstaining from all the SC votes.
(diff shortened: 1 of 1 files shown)


philiptaron (approved):
I agree that this is a pure refactor, as best as I can read it.

The middle commit message (3e8a681544f7b61a6c2efd713cb90121de18e4c1 at time of writing) contains "election election" which I believe is in error.

Ericson2314:
Thanks for the approval and noticing that! I've fixed the wording of that commit message.

doc/constitution.md
```diff
@@ -214,3 +217,6 @@ The reasons include but are not limited to unfitting conduct and unavailability,
 The SC is explicitly expected to be held to higher standards than applied to project communication in general.
 
 The removal can be justification for a special election where the removed person is not eligible to be a candidate.
+
+[special election]: #special-elections
+[initial election]: #special-elections
```
wolfgangwalther, line 226: This seems wrong.

doc/constitution.md
```diff
@@ -200,10 +200,14 @@ If it is impossible to find sufficient members to resign, all members with that
 
 #### Special Elections
 
-In the event of a resignation or other loss of a steering committee member (including a [removal for conduct](#removal-for-conduct)), a special election for that seat may be held if the SC deems it necessary, or the SC does not have half of the normal size.
+In the event of resignations or other loss of steering committee members (including [removals for conduct](#removal-for-conduct)), the SC would have less than all 7 seats filled.
```
wolfgangwalther, line 203: Changing from singular to plural in resignation, member and removal seems worse than before to me.

"Is this only meant for >1 or a single one, too?" This is clarified later, but no need to make it unclear in the first place.

doc/constitution.md
```diff
@@ -200,10 +200,14 @@ If it is impossible to find sufficient members to resign, all members with that
...
+
+When one or more seats are vacant, a special election may be held for those vacant seats.
 Otherwise, the missing seats are assumed to be abstaining from all the SC votes.
 A committee member elected in a special election will serve out the remainder of the term of the person they are replacing, regardless of the length of that remainder.
 
+A special election for vacant seats may be held if the SC deems it necessary, or the SC does not have half of the normal size.
```
wolfgangwalther, line 209: This sentence is unclear (but it was before already). If the SC does not have half the normal size, *may* a special election be held or *must* it be? I think the latter is intended.

(possibly out of scope here)

doc/constitution.md
Ericson2314: the case I am thinking of is something like:

1. After one member resigns, SC doesn't bother holding special elections
2. After second member resigns, SC holds special elections for both seats

The singular / edge-triggered wording makes it a bit unclear whether special elections are just for the most recently-vacated seat, or for all vacant seat.

doc/constitution.md
Ericson2314: I think it is in scope if we want it to be :). My view matches yours: it *must* be held.

Here's a suggestion
```suggestion
When a seat a vacated, a special election might be planned.
If the majority of the seats are still filled, the SC decides whether or hold a special election for vacant seats.
If only minority of seats are still filled, either a special election or the dissolution of the entire SC (which is always something that may be voted on) must happen.
```

PR #217, opened 2025-12-08
Give cafkafk access to calendar

tomberek:
Fixes: https://github.com/NixOS/org/issues/192

PR #218, opened 2025-12-09
CODEOWNERS: add @NixOS/steering as the fallback codeowner of every documentation file

philiptaron:
The situation of #217 with no merge happening because of incorrect paperwork -- even though it's just documenting a change that already happened -- is *ridiculous*. Let's fix that.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,35 +1,35 @@
 # See doc/org-repo.md
 
-/README.md @NixOS/org
-/CONTRIBUTING.md @NixOS/org
-/LICENSE @NixOS/org
+/README.md @NixOS/steering @NixOS/org
+/CONTRIBUTING.md @NixOS/steering @NixOS/org
+/LICENSE @NixOS/steering @NixOS/org
 
-/.gitignore @NixOS/org
-/.github/CODEOWNERS @NixOS/org
-/.github/workflows @NixOS/org
-/scripts @NixOS/org
-/npins @NixOS/org
-/default.nix @NixOS/org
-/shell.nix @NixOS/org
-/rulesets @NixOS/org
-/rulesets/nixpkgs @NixOS/nixpkgs-core
+/.gitignore @NixOS/steering @NixOS/org
+/.github/CODEOWNERS @NixOS/steering @NixOS/org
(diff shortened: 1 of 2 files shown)


mweinelt:
How about allowing the SC to bypass branch protection by ticking the box instead?

philiptaron:
That's a sensible change to make. I don't have access to make that change. Hence the ridiculousness circularity.

infinisil:
@philiptaron The idea of this repo is that it reflects the very distributed nature of who owns what. Code ownership should mirror who can actually affect the resources, because they are the ones who will have to implement changes to those resources. This all only works if the people with access actually maintain the documentation and keep it in sync with reality though.

The fact that we can't merge https://github.com/NixOS/org/pull/217 is merely a _symptom_ of the code owners of that file not maintaining it. The solution should be to change the code owners of that file, and the owners of the calendar with it.

Instead the main effect you get from adding the SC everywhere is that 7 more people get notifications for every PR in this repo, which robs everybody's attention, while diluting and centralising responsibility, and ignoring that the calendar and its docs is not being maintained by those with access to it.

philiptaron:
Why are you making it more difficult to do that?! Please click *some button* so that we can do that!

philiptaron:
I'm ticked off at your request to keep the docs up to date, then not clicking the button to merge the docs when they were updated. I don't care about CODEOWNERS files! I don't care about grain-of-rice-by-grain-of-rice permissions on files in this repo! I care about the documentation for NixOS governance being kept up to date, and because even the *steering committee* doesn't have rights to do that, *of course* it's not being kept up to date!

It should not require six different attention switches and private reach-outs and delays to do this. It should just have been `git push`.

infinisil:
I would've appreciated a more level-headed discussion, but okay, I turned off the code-owner-required check, so anybody with merge access can now merge any PR. I agree that that made some changes annoying, even though I liked the concept of it.

wolfgangwalther:
> I turned off the code-owner-required check, so anybody with merge access can now merge any PR

Thank you, that makes managing the branch protection rules easier as well, because I don't need an after-the-fact approval of *another* core team member, when I put up a PR myself - which was a bit annoying as well.

philiptaron:
Thanks @infinisil.

PR #221, opened 2025-12-10
org-repo: update to reflect the more relaxed merge rules

philiptaron:
See https://github.com/NixOS/org/pull/218 for the change.

I also did some copy editing, but no material changes.


infinisil:
Oh that slipped, thanks for the merge :)

PR #231, opened 2026-01-23
GitHub org owners need to add enterprise licenses

infinisil:
Ping @NixOS/org and @emilazy


mweinelt:
What does that even mean?

infinisil:
This is a little quirk of GitHub Enterprise: You can only invite new users if you have available licenses (which would usually cost). Licenses can only be added manually, there's no API endpoint, but we can add up to 300 licenses at once. I did this a couple times when we got enterprise, so we're currently at 4388 out of 6194 available. Every once in a while we just need to add some more before we run out again.

Lassulus (approved):

philiptaron (approved):

winterqt (approved):

PR #234, opened 2026-02-06
hetzner: update projects and infra acces

mweinelt:


infinisil (approved):
This is really great, thank you!

doc/hetzner.md
```diff
@@ -2,14 +2,40 @@
...
+- [@jfly](https://github.com/jfly) member
+- [@mweinelt](https://github.com/mweinelt) member
+- [@vcunat](https://github.com/vcunat) member
+
+### nixos-wiki
+
+The official wiki instance at [https://wiki.nixos.org].
+
+Project members:
+- [@JulienMalka](https://github.com/JulienMalka) admin
```
Mic92, line 29: Julian is no longer involved in this.

Mic92: I don't think @vcunat is there either. I can check though if he has access.

vcunat: I believe I do.  As a quick confirmation, I now logged into console.hetzner.com and saw `umbriel` in there.

Mic92: Julian indeed had access to the project. However he expressed before that he doesn't want to help administrating the wiki. I dropped him now from the hetzner project.

PR #239, opened 2026-03-02
Add freescout workflows module

infinisil:
Unfortunately only admins (and I'm the only one) can configure workflows right now. There's an option to allow any user to manage them, but it can't be changed due to https://cyberchaos.dev/e1mo/freescout-nix-flake/-/issues/3. And while we can adjust settings with Nix, I don't know the environment variable used by the workflows module to configure this permission.

### DIFF
--- a/doc/freescout.md
+++ b/doc/freescout.md
@@ -8,7 +8,7 @@ You can request a shared inbox for your team, or changes to an existing inbox by
 
 For a new shared inbox, make sure to include:
 - Team name (will be used in email footer):
-- Email address: <email-handle>@nixos.org
+- Email address: \<email-handle\>@nixos.org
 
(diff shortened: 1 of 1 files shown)


jfly (commented):
🚢

infinisil:
@jfly If you're interested in fixing this (and have some time), could you go to the freescout's directory in `/var/lib`, into the source of the Workflows module (it should be downloaded to a `Modules` or `public/modules` directory), and search for the environment variable linked to the string "Users are allowed to manage workflows"? :smile: 

Then we should be able to set it in https://github.com/NixOS/infra/blob/9944802491764a7d678bfc33e200f8ccca9f9b15/non-critical-infra/modules/mailserver/freescout.nix#L17 to make it work :D

PR #244, opened 2026-04-20
Revise NixCon 2026 sponsorship details (only organisers version)

infinisil:
> [!Note]
> Alternative to https://github.com/NixOS/org/pull/251
>  This version does _not_ allow volunteers with responsibilities to discuss sponsors before they're approved/rejected

It's not ideal how the [2026 sponsorship tiers](https://2026.nixcon.org/#sponsors) were changed without updating the document here. I'm also not sure how the changes were decided, but I don't think they need SC/board approval, so fine by me :)

The more important change here is a simplified sponsorship approval process with much lower bureaucracy compared to last years.

@NixOS/steering Please majority approve
@ners @john-rodewald @Lassulus @Ra33it0 Please approve

PR #251, opened 2026-04-27
Revise NixCon 2026 sponsorship details (including volunteers version)

infinisil:
> [!Note]
> Alternative to https://github.com/NixOS/org/pull/244
> This version allows all volunteers with responsibilities to discuss sponsors before they're approved/rejected

Would need approval by the @NixOS/steering and NixCon organisers (@Lassulus @Ra33it0 @ners @john-rodewald)


ners (changes requested):
We discussed in today's orga meeting and decided to go with https://github.com/NixOS/org/pull/244 instead. This does not mean that volunteers are not part of the discussion (after all, they may decide to no longer volunteer if some decision is really not to their liking), but we are trying to keep the process lean this year.

As I do not have permission to close the PR, I've commented with "Request changes".

tomberek:
Closing for #244 .

Issue #242, opened 2026-03-26
Access to the NixOS.org google calendar


Mic92:
It's a bit of a pain just now when we manage teams and want to change calendar events and we always have to reach out for one of the few folks that can edit.
Since I am on 4 teams (hydra, nix, infra, hardware) it might make sense to me to also be able to manage the google calendar. My gmail address is jthalheim@gmail.com
Thanks for considerations.

winterqt:
cc @tomberek @cafkafk @infinisil

infinisil:
Please PR to https://github.com/NixOS/org/blob/main/doc/calendar.md

philiptaron:
I support this!

PR #264, opened 2026-06-03
Clarify ownership of google calendar and give access to @Mic92

infinisil:
Improve the docs a bit, there's a more elevated owner permission

Give access to @Mic92 as requested in #242


mweinelt:
Would it be helpful to annotate, why someone has received permission to edit? E.g. in my case infra calls.

infinisil:
Unless we have a problem that would be fixed by this (which I don't think there is), probably not worth bothering

mweinelt:
The purpose would be gc roots. :wink:

infinisil:
Makes sense, done!

tomberek:
Done

PR #270, opened 2026-06-15
Proposal: Grant program and purpose-bound funding

infinisil:
In recent months, primarily the NixOS Foundation board members, me and @ra33it0, have been drafting this proposal to fill some gaps in the NixOS Foundation's funding model. Note that this is not yet fully board or SC approved, which we'll need to get before this is implemented.

### Brief summary

Motivation:
- Enable companies to finance Nix ecosystem improvements in a more streamlined and transparent way
- Enable contributors to get paid for working on Nix ecosystem improvements in a more streamlined way
- Provide a way to connect funding partners to contributing individuals, e.g. for when:
  - Individuals don't want to be employed/contracted directly by a company, and don't want to be limited by receiving funding from only one company
  - Companies don't want to employ/contract an individual directly, and want to provide funding in a more flexible way to multiple individuals
  - Companies want a trusted third-party to hold the funds until work is complete
- Strengthen the ecosystem by providing funding resources to contributors using general NixOS Foundation donation profits not tied to any companies

There are two inter-related parts to this proposal:
- Grant program: A way for contributors to propose ideas and get paid by the NixOS Foundation for delivering milestones. The funds can come from both general-purpose donations and purpose-bound funding.
- Purpose-bound funding: A way for companies to provide funding for specific grants proposals or team initiatives.

### Current status

- [x] Revised after feedback from board and SC members
- [ ] Revised after feedback from official teams
  - Pinging these teams for feedback (somewhat arbitrary subset from [the homepage](https://nixos.org/community/#teams)): @NixOS/nixpkgs-core, @NixOS/security @NixOS/infra @NixOS/marketing-team @NixOS/moderation @NixOS/cuda-maintainers @nixos/nix-team @NixOS/documentation-team @NixOS/nix-formatting @NixOS/ngi @NixOS/nixos-hardware @NixOS/hydra-maintainers 
- [x] Shared with the wider community: https://discourse.nixos.org/t/nixos-foundation-draft-proposal-for-grant-program-and-purpose-bound-funding/79732
- [x] Shared with companies
- [ ] Approved by board and SC

If we need to add more context behind individual parts, we'll add this on demand.


proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+A grant wave is 3 months long, comprised of two phases, and executed by the board or someone delegated:
+
+- Wave start: Widely announce the new wave, including:
+  - Decided by the board: how much funds (and how they're ear-marked, if at all) are available to be allocated. This generally includes part of the income and ear-marked funds.
+  - Decided by the SC: The agenda that will be prioritised for this wave
+- Proposal phase (1.5 months): Anybody can write a grant proposal by submitting a PR to a repository.
+  - Proposals should primarily include the team to do the work, the milestones, timeline, the amount needed and the recipient of the funds (can be individuals or companies)
+    - Proposals may also be for non-official projects
+  - A team can be a single individual or multiple individuals with one of them acting as project manager
```
fricklerhandwerk: No need to specify what a team is if we don't introduce the term in the first place

```suggestion
  - Proposals should primarily specify the people doing the work, the milestones, timeline, the amount needed and the recipient of the funds (can be individuals or companies)
    - Proposals may also be for non-official projects
```

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,95 @@
...
+- Wave midpoint: Close proposal submissions, announce the feedback phase and reach out to:
+  - Various stakeholders for feedback, including teams or individuals
```
fricklerhandwerk:
```suggestion
  - Various stakeholders for feedback, including community teams or individual contributors
```

Is that what you mean?

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,95 @@
...
+  - The board ensures standard legal requirements
+  - Standard rejection message that hints towards alternative funding partners
+
+### Milestone evaluation
+
+After grant proposals are kicked off, the applicants are expected to do the work and follow up after completing each milestone with detailed reports submitted as a PRs to the evaluation repo. An appointed evaluator will evaluate the work, request changes if necessary, and give approval if satisfactory, in which case the applicants will be paid.
+
+- The board/SC trusts the evaluators to do the evaluation properly
+- Evaluators are required to submit an evaluation report including deliverables, evaluation notes and approval decision
+- Evaluators get 5% of the milestone amount
```
fricklerhandwerk: This section underspecifies how and when evaluators are appointed. Not saying you need to nail it down now, but since evaluation (or call it "review") is a well-known and somewhat well-understood bottleneck on delivery, it probably needs to be planned ahead of time. I'd even say it needs to be priced right into the original proposal. Which is why it sticks out to me that you fix the review overhead by fiat; I argue this will be highly domain- or even case-specific, because reviewer qualification and time isn't fungible.

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+![](./waves.svg)
+
+([source](https://excalidraw.com/#json=rRutoi8nEcpjv1UIj3X87,VCPIeHZP_6ouG0U7tWQxMg), make sure to update the source link when changing anything)
+
+## Initiative funding
+
+- Establish a process for going into contracts with companies to supply funding for specific purposes, with a guarantee that the funding (without an additional 10% management overhead) will be used for the stated purpose, and that if not all of the funding could be used as such, the rest is repurposed according to the funder companies wishes.
+  - There is a 10% management overhead that will not be allocated to the initiative funding.
```
fricklerhandwerk: Is there a particular calculation behind this number? Because from public funding for things like research, this order of magnitude seems to only fly if there already exists sufficient baseline funding independent of such programs. For comparison, Horizon grants give 25% "indirect costs" allowance for operational expenses unrelated to the immediate program goals, and this seems like more in line with the effort of setting up and maintaining a procedure like it's outlined in this document.

One can argue that some effort can be expected to be volunteered for those purposes, but if the goal is to have not just streamlined but *reliable* operations for passing money through to who is doing the work, that part of the process should better be equipped with the resources to keep those responsible for it around.

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,95 @@
...
+  - The minimum required funding is 1000 EUR (+10% overhead, excl. VAT).
```
fricklerhandwerk: Is there a calculation or bet you're making behind this number? One could argue that 10% is enough up from a certain threshold, but 10 sponsors of that size will amount to *a couple of hours* of preparation and upkeep. So unless sponsors are fighting to spend their money on NixOS stuff already, or I'm missing some fact about how to do things extremely swiftly without having done that sort of stuff for months or years (the argument above about keeping people around for a long time), off the cuff this seems unlikely to be self-sustaining. Or is it just about making the entry price look reasonable?

fricklerhandwerk (approved):
I think this is a very good idea overall, and the rough structure seems sound. It's a miniature version of what the European Commission does. If the costs and benefits of doing such a thing are well understood, this can be a good mechanism for coordinating action that would otherwise not happen.

The only thing I'd caution against is assuming that the capacity to operate such a scheme will magically appear out of the blue. This needs considerable up-front investment from everyone involved, and quite a bit of time for building experience and routines, including those to trust and delegate and exercise the control mechanisms outlined here. What may help bootstrapping it is managing expectations with all parties: Initially we can expect that sponsors won't get a good bang for their buck, implementors and reviewers won't get good rates, organisers will do serious work on their free time, balls will get dropped, and there won't be too many exciting success stories for the general audience of users or maintainers. But that's how things go and will be fine if it's approached as a learning process aimed at getting it right eventually.

cafkafk (approved):
I already gave feedback in the call yesterday, so I've approved it. I'd still like to see more teams feedback of course

friedow (approved):

proposals/grants-and-initiatives/flow.svg
YorikSar, line 1: There should be either an arrow from "OpenCollective" to "generic funds" for the fiscal host or a clarification that this only concerns specific projects.

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,95 @@
...
+  - The [SC's evaluation criteria and process](#sc-evaluation-criteria-and-process) have to be met
+  - The board ensures standard legal requirements
+- Create a page on the website that showcases a list of efforts that are looking for funding
+  - This includes team-submitted initiatives and grant proposals from the current feedback phase
+    - Official teams can submit initiatives by creating a PR against the nixos-homepage repo, which is reviewed by the board/SC before merged
+  - There is a contact form for companies to reach out to initiate the process of submitting funding for either a listed effort, or a custom one
+  - A rough mockup might look like this, but there should also be a clickable email link as an alternative:
```
YorikSar, line 80: Should initiatives.svg be here?

proposals/grants-and-initiatives/initiatives.svg
YorikSar, line 1: Asterisks by Name and Email fields mean a different thing than one at the Total.

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,95 @@
...
+## Grant program
+
+Create a page on the website to explain this.
+
+### Proposal evaluation waves
+
+A grant wave is 3 months long, comprised of two phases, and executed by the board or someone delegated:
+
+- Wave start: Widely announce the new wave, including:
+  - Decided by the board: how much funds (and how they're ear-marked, if at all) are available to be allocated. This generally includes part of the income and ear-marked funds.
```
YorikSar, line 31: I think it would be better not to use "decided" with the board. Maybe "calculated" or "tabulated"?

proposals/grants-and-initiatives/README.md
YorikSar: Project manager is required here, but their role is not defined.

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,95 @@
...
+  - Decided by the SC: The agenda that will be prioritised for this wave
+- Proposal phase (1.5 months): Anybody can write a grant proposal by submitting a PR to a repository.
```
YorikSar, line 34: It's unclear if we need to submit new proposal for each wave, or the repo is open for PR always and SC somehow picks from all open PRs.

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,95 @@
...
+    - Proposals may also be for non-official projects
+  - A team can be a single individual or multiple individuals with one of them acting as project manager
+- Wave midpoint: Close proposal submissions, announce the feedback phase and reach out to:
+  - Various stakeholders for feedback, including teams or individuals
+  - Companies for the opportunity to submit extra funding for specific proposals (via initiative funding)
+    - Also mention team initiatives, same as on the website
+- Feedback phase (1.5 months): Collection of feedback and funds
+  - Feedback from stakeholders can be discussed and acted upon
+  - Sign initiative funding contracts with funding companies wanting to submit funds for specific proposals
+- Wave end: SC decides which proposals get funded, while proritising the agenda and satisfying the [SC's evaluation criteria and process](#sc-evaluation-criteria-and-process), which are then merged to signal that the proposal is accepted for funding and the work can start
```
YorikSar, line 46:
```suggestion
- Wave end: SC decides which proposals get funded, while prioritising the agenda and satisfying the [SC's evaluation criteria and process](#sc-evaluation-criteria-and-process), which are then merged to signal that the proposal is accepted for funding and the work can start
```

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,95 @@
...
+  - The board ensures standard legal requirements
+  - Standard rejection message that hints towards alternative funding partners
```
YorikSar: No action in this point, unclear.

lisanna-dettwyler (approved):
Nothing especially stands out to me as problematic, but I think @fricklerhandwerk made excellent points.

SomeoneSerge:
Hey-hey, apologies for putting the "thumb down" and going radio-silent! It's a very important and complex subject, so it's not easy picking words to find the balance between actually communicating information and keeping the debate constructive...

First off, thank you for the effort of putting your vision together and into words. I believe it's hard, it's hard work and genuine. And I'm grateful for it even when I disagree on any particular account.

The proposal. I feel concerned about this, along several lines of thinking, of which the most important are these three questions:

1. Who are the "companies" you target with this proposal?
2. How does this proposed system help maintainers negotiate for higher prices for their work, instead of helping companies always find the cheaper labour and create incentive for dumping? The former being the zero-order approximation of what we need for the project to succeed.
3. What is the CoI story of this proposal?

Now to elaborate. We haven't discussed any of this face-to-face yet, but you're a Board member and in close touch with Yuriy, so I can't imagine you haven't seen my own write-up on the subject (not linking because too dramatic) or heard what I've told Ron and Jonas, which makes me feel like I'm repeating myself, because these are all the same questions I've been asking for feedback on for a year, and I feel like you've ignored them.

#### "Which Companies?"

I am asking (1) "which companies", because the proposal reads as if completely out of the context of the preceding year's hard talks and negotiations. If this proposal is in any way at all connected to this past year's work, and all of the sacrifices that have been made to grant the Foundation a seat at the table as a pretext for everyone to start building a framework for future collaboration with the very concrete and non-abstract companies, then..

I must say I find that this proposal **discusses everything that is unimportant, irrelevant, and nothing that is urgently begging solutions**. I fear we might be still trying to achieve very different and deeply incompatible things, the first give-away being your use of the word "donations", and that other readers compare this effort to EC grants corroborates this. After all, we have just seen the demonstration of how companies' boards abhor the notion of a donation, almost as much as they cringe at the notion of working with a non-profit, almost as much as they hate working with a _European_ non-profit. This is the one and only matter on which we-maintainers and they-corporations actually do not have to disagree, because the legal reality is that it is easier to move bigger quantity of money as service fees than as donations anyway, but we're now deliberately trying to make ourselves worse off...

The proposal conjures up a number of compliance rituals and rain-dances nobody has ever asked for during this past year:

- we haven't run into any situation that'd suggest we need designated "grant waves", so why quantization?
- we haven't yet had a need to fix a "wave's" duration, why?
- we haven't had a need to have designated "proposal phases" that have distinct start and end points, so why?
- we haven't actually run into a concrete situation where we need the Board's or the SC's involvement, so what is the motivation for introducing these additional synchronization steps and additional parties? An incredible amount of rain-dancing already, and we haven't even start talking about any actual work!
- "SC decides which proposals get funded", nitpick but well not really it don't? This entire proposal is a match-making system (tuned to minimize the amounts of money spent by the "sponsors" or reaching the maintainers), and once a match is found it'll probably get a go whether with or without the Foundation as a proxy. I don't disagree with the idea that it's actually nice to have the SC in the conversation, but the whole unpnaid ceremonial compliance dancing thing... le sigh
- still the same question for introducing "evaluators" and for pre-defining the reporting rules: why give up and accept the worse terms before the fight even began?
- ...now that we've introduced "evaluators" (nobody asked), why do we pre-define their share?
- a lot more importantly, why do we pre-define the share of the Foundation and WHY DO WE SET IT SO LOOOOOOOOOOOOOOOOOOW? I mean, we made this this blunder with the first round of the anti-cuda funding, okay, there was too much happening, too much stress, but now that we're not in a rush can we finally get this right? <details>This number can be any number, because there's so much inter-dependency between everyone's work in Nixpkgs&co that even 99% is frankly speaking justifiable. "Management overhead" is an expense article for the project, and it's basically pre-approved. It's the first expense that automatically happens and doesn't require any justification or any MS Teams meetings. Once this "management overhead" part of the project budget is spent, it is finally clear of obligations to the corporations, it becomes the Foundation's own money. This means that _then finally_ it can be reinvested in something actually useful to maintainers and to your general human, instead of on the subset of easily-defined projects that very clearly generate company profits.</details>

#### "Help or harm?"

The proposal is still missing a reflection on the most important question: why try put the Foundation into this position in the first place? I find it self-evident that, materially speaking, the best for a maintainer is to have service contracts directly with their customers. As I wrote in my piece, the only reason ever to consider setting up a proxy, such as the Foundation is, is when there are multiple maintainers offering the same service (project), i.e. when we find ourselves competing. Then there's also the logistics matter, which is that it's cheaper procurement- and taxation-wise for corporations to only deal with the single Trust-entity proxy. Even then the Foundation is the second-worst choice, because these said maintainers are not owners of the Foundation. There must be some Guild-like configurations in which the Foundation can be beneficial instead of harmful for maintainers, where the Foundation helps the match-making and helps negotiate for higher prices, helps to sign direct contracts and establish direct communication channels between maintainers and corporations, but this proposal isn't offering any of that, nor have I come up with anything tangible myself.

WARNING: Try to read the following paragraph not as a complaint, but as an account of my own mistakes and failures. A "system", e.g. as proposed by the PR, could be in place to exacerbate or compensate such failures. My argument is that the PR proposes that we take a small system which made things worse and harder specifically for me in the concrete scenario, and that we scale it up to all of Nixpkgs. My expectation is that it would then make things worse at scale.

I can only offer counter-proposals in the very narrow and counter-factual context of the cuda effort, with the hindsight of how it's gone terribly awry. The counter-proposal for what the Nixpkgs CUDA effort _should_ have been, and what it can't be anymore because the landscape got so much more complex, is, very bluntly and self-indulgently, that it very obviously should have gone through my company, and my company should have donated to the Foundation, no strings attached.<details> (A) "Why hasn't this happened?" I haven't pushed hard enough for this during Autumn-Spring because I was busy trying to keep my head above the water, and especially demotivated after MODUS coming into the picture, and even more so after the initial co-funding proposal. (B) "Why would it have been right?". Because the direct relationship (assuming it's the most beneficial to maintainer) is the default way to go, except when we can't, and in Nixpkgs CUDA's unique situation we can. In case of Nixpkgs CUDA Team as it is today, we haven't any hard conflicts of interests, speaking materially and financially. Even today, I and Daniel are the only people on the team willing or ready to invoice for CUDA stuff, and Daniel's is, for the moment, a completely nominal fee so it's not a conflict... I was hoping this would no longer be the case by now! It's been an entire year since I first asked Ron for an opinion on the COI problem, and this has been another reason I haven't pushed decisively against the previous Foundation Deal. We could have been doing work and making progress during this one year, I could have averted all of my external crises, put out the fires, plugged the holes, and focused on the CUDA bullshit. I could have set up all of the infra one year ago, as planned. We could have gone incrementally and started with cloud compute were it not for the compromised trust and for the insecure position created by the lack of a margin. But we keep playing games and trying to save corporate money, which the corporations don't even want saved. (C) "Does that scale to all of Nixpkgs?". I don't know, not in any obvious way. Generally, and even long-term with cuda, there will be conflicts of interests, so we'll need proxies, and these proxies better be collectively owned. (D) The Foundation could have been helpful in this situation. The Foundation can help me steer the conversation towards the direct relationship, can help me brainstorm how to make this an SLA story, can have help negotiate.</details>Essentially, the Foundation could hypothetically function as a "guild" or a "pricing consultancy", at least in this very specific and easy case, if not in general. I had said I fear the Foundation becoming a "trust-entity" instead, and your proposal, I think, is bound to make it exactly that...

#### COIs behind the proposal

It really would help the conversation a whole lot if you included a reflection on what Sylvain-the Board member thinks of how FOSS and/or Nixpkgs funding should work, what Sylvain-MODUS employee, and what Sylvain-Nixpkgs maintainer, all three being essentially conflicting roles. For example, I hope that I've been clear enough that I have a purely materialistic interest in the CUDA stuff, that my well-being at various times has and will have depended on how this effort goes, and ergo I have an obvious conflict of interests.

#### To conclude...

I don't see why we need this system or any of the low-level details fixed in the proposal, and I see too many ways in which this system is harmful and counter-productive. I still hold my so far unchanged opinion, that the most reasonable way in which "big companies" are able to disburse of money is through service fees, maybe with the exception of special companies which spend money on lobbying, - in any case not through "donations". There are companies that are willing to invest in the ecosystem, and they can invest a "rounding error on the balance sheet" amount of money and it would still be more than enough for us to focus on productive work, be able to satisfy their wishes, reinvest the remaining margin in the actual useful stuff, and never ever have to LARP the European Commission or NLNet. The last two are concerned with a much harder task, which is justifying the expenditure of a very scarce money held by entities that really don't want it spent, especially on something like FOSS. The entities that I assume this proposal is targeting are nothing like that, they are directly benefiting from our ecosystem, and our ecosystem's prosperity impacts their profits. So let's focus on how we have a unique chance to change how FOSS is funded if only we can figure out the exact steps and the exact shape?

As always, "this would have been shorter", but it's already been a week and nobody got the time to wait for me to filter my messy thoughts

UPD1: To put things to scale, if the Foundation were on a course to collect a total of $1bil for a year in "donations" and "funding", I'd not say a word. The way things currently are going, I don't see us reaching 10% of that bar.
UPD2: And again, don't get me wrong, I failed plenty, and I'll fail more, and I'll own my failures. But we're having a discussion about how the Foundation can be of help and the Foundation can be harmful, and here the concrete examples of both, as plain as I could frame them.

fricklerhandwerk:
Thank you @SomeoneSerge  for bringing this up with such a detailed criticism. This sharpens my view as well, to the point I would suggest to @infinisil an editorial pass to deemphasize or leave open the minutiae (also to give some artistic liberty to implementers) and more clearly carve out the incentive structure and economic model. I think it's already in there but will hugely benefit from sharpening. Or in a nutshell: make it more of a guideline rather than a set of instructions, at least for now.

<details>

<summary>Detailed reply to @SomeoneSerge's comment</summary>

While I don't agree with the criticism of low-level details (a procedural framework absolutely has benefits, the EC and NLnet absolutely want their money spent, it's their business and vital interest, etc.), I can imagine your frustration with the current realities very well, and I agree with the gist that what's missing is the big-picture story of why doing such a scheme is a good idea and, in particular, why, how exactly and for whom it's worth the extra effort. Part of making that more legible would include leaving open the moving parts such as margins and cost structure, as I also mentioned in my previous comment.


I was reading the proposal favorably though, seeing it as strongly implied that the foundation would be a matchmaker exactly because orgs often can't easily "just hire" individuals, and individuals often don't have the capacity to coordinate large deals. Whether you call that donation or sponsoring doesn't seem to matter for the spirit of the law here, at least for me since I see the SC involvement part of the proposal as the place where CoI is negotiated.

</details>

SomeoneSerge:
> because orgs often can't easily "just hire" individuals

Well we don't want them "just hiring" individuals, at least if "hiring" means employment, because then they can't negotiate for terms that respect COIs

> and individuals often don't have the capacity to coordinate large deals

A problem very easily solved with enough resources. There, another prospect for the Foundation to be a provider of help instead of the cheaper competitor to the maintainer:)

EDIT:

> the EC and NLnet absolutely want their money spent

The NLNet sure does. But how much effort it takes for NLNet to get their money? Wasn't there a public outcry just recently about the EC trying to cut NGI off funding?

SomeoneSerge:
Let's amuse ourselves and ignore the logistics cost of maintaining `N * M` contracts, one could imagine a match-making system where the money does not go through the Foundation, but instead through the maintainer-owned entity to the Foundation. The Foundation then is a facilitator, the observer, a participant in the bargaining, may oversee the private companies' contract terms and ask for a cut. So is the SC. As a fallback measure, when a group of maintainers doesn't have an entity set-up yet could the Foundation resort to overtaking the deal and signing the contracts instead of the maintainer. The Foundation could eventually learn to provide legal and consulting services, helping maintainers to set up their entities. The Foundation could reinvest its share in financing the maintainers during the very slow bargaining and procurement periods, thus again increasing maintainers' capacity for negotiation instead of diminishing it as the proposal does now.

There, the Foundation is no longer a competitor, but a partner. Can we adjust this fairy-tale so it scales to the project of our size?

fricklerhandwerk:
Do I read correctly that what you're proposing at the core is to model this (as you said) more like a guild or worker's association or cooperative with shared ownership, with resource distribution schemes developed accordingly, instead of as a central planning office under supervision by an elected body? If so, I think this is perfectly valid, but also both models converge to roughly the same structures in practice, in particular a group of representatives focusing on the coordination work. It seems the fundamental difference between the two is the direction of money flow and its consequences, primarily in terms of information asymmetry:

If contributors close their own deals, they need a good reason to pay the foundation a cut. Currently companies in the ecosystem already close their own deals, but typically pay their cut at their own discretion and mostly in-kind or in the form of NixCon sponsorship, which is very different in nature than free-to-use-money. The incentive is to keep the cut low and the foundation out of the loop, which results in the current situation of the one entity responsible for coordination being resource-starved. Contributors could just not disclose their deals, and that is typically the case, and it's mostly not evident who paid for what work.

If the foundation closes the deals, they can just keep the cut, enough to sustain operations according to statutes. The oversight is supposed to make sure that things are reasonable. This of course concentrates power, but that's kind of what the elections are for. And some degree of centralisation is the trade-off you have to make to get more efficient coordination. The information asymmetry here is that a central body may not disclose market insights such as potential deals or the competition landscape.

Pooling risk (as in your example of paying though slumps) seems to be independent from that, the foundation as an entity can in principle offer this service in both setups, but that seems out of scope here.

It seems the real questions here are whether this coordination problem is serious enough it needs a central entity to mitigate (I'm convinced that is the case), what specific service that entity would offer and at what price (this document makes a proposal), and who owns or benefits from what (currently underspecified, this would be important for me to add to this document).

I've participated in both models, and I'm convinced we currently need more of the second, and not even necessarily less of the first. Both already coexist, we just need to balance those interests and handle the dynamics.

infinisil:
@fricklerhandwerk @SomeoneSerge Thanks for the feedback, we'll consider it for the next revision. To leave more room for other people I'd like to request you continue your discussion among yourself in an out-of-band channel for now.

proposals/grants-and-initiatives/README.md
infinisil: Resolved by https://github.com/NixOS/org/pull/270/changes/4b5b23a98038994bf1318c29e4805c98ec4779b8

proposals/grants-and-initiatives/README.md
infinisil: The reason for €1000 minimum is because we want to allow smaller companies to contribute as well. As a comparison, for NixCon 2025 there were 6 monetary sponsors for the €1024 tier. And we expect the initiative funding overhead to be low, but are also open to change it in the next revision if necessary.

I'll answer the question about the 10% in https://github.com/NixOS/org/pull/270#discussion_r3413437849

proposals/grants-and-initiatives/README.md
infinisil: Some reasons:
- We expect way less overhead, we made an estimation and expect to get by with the 10%
- We want to start small, and stay flexible to adjust it later if necessary

proposals/grants-and-initiatives/README.md
infinisil: https://github.com/NixOS/org/commit/2f0769aba3885f8363308631bc100cc0816d4212

proposals/grants-and-initiatives/README.md
infinisil: https://github.com/NixOS/org/commit/1351ef745bde78e3b8397d1b2fad53f5a429a0c8

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,120 @@
...
+  - There is a contact form for companies to reach out to initiate the process of submitting purpose-bound funding for either a listed opportunity, or a custom purpose, see below for a rough mockup
+  - An email to reach out to should be provided as an alternative
+- If there's a lot of interest by community individuals, the board can consider creating an OpenCollective crowd-funding campaign for team initiatives
+  - As opposed to purpose-bound funding, such crowd-funding campaigns are a donation without a legal guarantee of the funding being used for the purpose
+
+![](./initiatives.svg)
+
+([source](https://excalidraw.com/#json=ibw0fNwkkGuL1QGhAhVIx,cJ-7--yHQVGMEH3Wfcj08A), make sure to update the source link when changing anything)
+
+## SC evaluation criteria and process
```
tomberek, line 100: Most of this section goes over reasons to reject; harms and their various categories. Recommand having at least a section of reasons to accept; positive reasons about evaluating benefits of a proposal in terms of accomplishing the Foundation's goals or being in line with the Values.

Otherwise. Looks good overall!

tomberek (commented):
Looks good, just would like to balance out the criteria with positive reasons as well.

infinisil:
Me and @Ra33it0 have now made a number of revisions based on the feedback so far. I encourage everybody to re-read the proposal in full, but a notable point is that we've strengthened the _team initiatives_ concept in comparison to grant proposals: Purpose-bound funding for grant proposals now has an additional 20% maintenance and 5% evaluation charge (in addition to 10% for management), while purpose-bound funding for team initiatives only has the 10% for management. Furthermore, funding for team initiatives is encouraged to be recurring instead of one-time.

In general I'd also like to point out that this proposal is _not_ perfect, and _will_ need changes over time. While we try to be as forward-looking as possible, at some point it's better to try something out and iterate on it with real-world data, instead of never getting it out of the discussion phase.

proposals/grants-and-initiatives/README.md
cafkafk: I’d think about it in terms of which criteria are actually useful for applicants and for the selection process.

From an applicant’s perspective, it helps to understand what would make an application ineligible or clearly out of scope. That avoids unnecessary effort, makes the process feel more approachable, and gives people a straightforward way to assess whether it makes sense to apply.

I would be more cautious about adding criteria for what a "good" application looks like. Once you start listing positive criteria, anything not on that list can easily be read as less relevant or less desirable, even if that was never the intention. This is a dynamic seen in hiring, for example, where people often self-select out when they do not see themselves reflected in the stated criteria.

Not having positive criteria gives the selection process more room to recognize strong applications on their own merits, including ones that may be good in ways we did not anticipate. The criteria here are useful as a way of defining the boundaries of the process, not as an exhaustive description of what we might value. I think a fuller discussion of what "good" can look like would be a separate document entirely and would be misplaced here.

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
+# Proposal: Grant program and purpose-bound funding
+
+Alternative title: Filling funding flow gaps
+
+## Motivation
+
+- Enable companies to finance Nix ecosystem improvements in a more streamlined and transparent way
+- Enable contributors to get paid for working on Nix ecosystem improvements in a more streamlined way
+- Provide a way to connect funding partners to contributing individuals, e.g. for when:
+  - Individuals don't want to be employed/contracted directly by a company, and don't want to be limited by receiving funding from only one company
```
SomeoneSerge, line 10: Do such individuals exist? Why should the Foundation cater to such individuals, if there were any, instead of helping individuals to get direct contracts?

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
...
+  - Companies don't want to employ/contract an individual directly, and want to provide funding in a more flexible way to multiple individuals
```
SomeoneSerge, line 11:
```suggestion
- Help contributors find customers and secure commercial contracts, or procure funding for otherwise "unmarketable" work.
```

Why should the Foundation help companies avoid direct contracts or create hidden employment relationships?

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
...
+  - Companies want a trusted third-party to hold the funds until work is complete
```
SomeoneSerge, line 12: Why should the Foundation help Companies in their anti-competition efforts?

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
...
+- Strengthen the ecosystem by providing funding resources to contributors using general NixOS Foundation donation profits not tied to any companies
+
+# Detailed Design
+
+There are two inter-related parts to this proposal:
+
+- Grant program: A way for contributors to propose ideas and get paid by the NixOS Foundation for delivering milestones. The funds can come from both general-purpose donations and purpose-bound funding.
+- Purpose-bound funding: A way for companies to provide funding for specific grants proposals or team initiatives.
+
+These programs should incorporate messaging to encourage contributing individuals to seek employment contracts or contracting agreements with funding partners.
```
SomeoneSerge, line 22:
```suggestion
These programs should incorporate messaging to encourage contributing individuals to seek  contracting agreements with funding partners.
```

...but I can't claim I understood the intent of the paragraph in the first place

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+![](./flow.svg)
+
+([source](https://excalidraw.com/#json=ZrF5CEDeZp5squJcqxXu_,LzsWVJZ_ZQXvCuoIPr9iwA), make sure to update the source link when changing anything)
+
+## Grant program
+
+Create a page on the website to explain this.
+
+### Proposal evaluation waves
```
SomeoneSerge, line 32: Already asked in the previous round, but why "waves"? Why not a continuous call?

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
...
+  - The 5% is budgeted internally by the foundation
+
+![](./waves.svg)
+
+([source](https://excalidraw.com/#json=rRutoi8nEcpjv1UIj3X87,VCPIeHZP_6ouG0U7tWQxMg), make sure to update the source link when changing anything)
+
+## Purpose-bound funding
+
+- Establish a process for going into contracts with companies to supply funding for specific purposes, with a guarantee that part of the funding will be used for the stated purpose, and that what could not be used as such is repurposed according to the funder companies wishes.
+  - The minimum required funding is 1000 EUR (plus management/maintenance/evaluation budget, excl. VAT).
```
SomeoneSerge: Please let's omit all numbers that aren't the order of $10^6..10^9$ from the document (applies to the diagrams too)

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
+# Proposal: Grant program and purpose-bound funding
+
+Alternative title: Filling funding flow gaps
+
+## Motivation
+
```
SomeoneSerge, line 6:
```suggestion

- Procure funding for experimental projects, fundamental research, and the "invisible work" that is otherwise hard to market.
- Provide contributors and their projects with interim funding during sponsor search or bargaining periods.
```


...here is one (and, fwiw, only) use-case for a "grant system" that I currently see. All of the work that is too removed from "profit-making" for the market to show interest in, maybe projects whose time hasn't come yet. Running a mainline kernel on nokia 3310 or whatever weird stuff Matthew does...

EDIT: added the 2nd, a variation of the 1st

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
...
+  - Companies don't want to employ/contract an individual directly, and want to provide funding in a more flexible way to multiple individuals
+  - Companies want a trusted third-party to hold the funds until work is complete
+- Strengthen the ecosystem by providing funding resources to contributors using general NixOS Foundation donation profits not tied to any companies
+
+# Detailed Design
+
+There are two inter-related parts to this proposal:
+
+- Grant program: A way for contributors to propose ideas and get paid by the NixOS Foundation for delivering milestones. The funds can come from both general-purpose donations and purpose-bound funding.
+- Purpose-bound funding: A way for companies to provide funding for specific grants proposals or team initiatives.
```
SomeoneSerge, line 20: I'm getting the impression that "purpose-bound funding" is expected to be the main source of cash for the grant system, and I really hope that's not the case? Anything that is "purpose-bound" is probably clear-cut enough to be a commercial contract between for-profit entities, but then what the Foundation should do is collect a sort of a "tax" (no strings attached) from such commercial fees in exchange for facilitating such deals and helping companies and contributors connect...

SomeoneSerge (commented):
I only briefly skimmed the document. I think I have the same principal question as before: what are some of the real-world efforts we're trying to accommodate here, so it's possible to evaluate the proposal against them?

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Proposal evaluation waves
+
+A grant wave is 3 months long, comprised of two phases, and executed by the board or someone delegated:
+
+- Wave start: Widely announce the new wave, including:
+  - Decided by the board: how much funds (and how they're ear-marked, if at all) are available to be allocated. This generally includes part of the income and ear-marked funds.
+    - In addition to the funding directly available to the grant proposals, internally an extra 5% is budgeted for grant evaluators
+  - Decided by the SC: The agenda that will be prioritised for this wave
+- Proposal phase (1.5 months): Anybody can write a grant proposal by submitting a PR to a repository.
```
SomeoneSerge, line 40: Should be possible to have private submissions until they're approved

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
...
+  - Proposals should primarily include the team (1+ persons, 1 contact person) to do the work, the milestones, timeline, the amount needed and the recipient of the funds (must be a company, can be self-proprietorship)
+    - Proposals may also be for non-official projects
+    - Timelines can be arbitrarily long and have arbitrarily many milestones in them
+  - Teams and funding partners are encouraged to directly synchronise with each other about the scope of the work and grant amount. Ideally the team submits something to the grant program that the company then submits funding for using the purpose-bound funding mechanism
```
SomeoneSerge, line 44: > Ideally the team submits something to the grant program that the company then submits funding for using the purpose-bound funding mechanism

Something I'd very specifically recommend discouraging! On the very rare occasion of a "double coincidence of wishes", the Foundation should be helping the team create an upward pressure for the price instead of fixing it low...

proposals/grants-and-initiatives/README.md
fricklerhandwerk: I agree that the focus should be for the foundation to provide a service to the community, if nothing else then merely to keep its non-profit status. It is easy to see such a scheme as a match making mechanism, but how is that not an agency?

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
+# Proposal: Grant program and purpose-bound funding
+
+Alternative title: Filling funding flow gaps
+
+## Motivation
```
fricklerhandwerk, line 5: The following are goals, not motivations. The document doesn't state *why* these are goals to begin with.

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,116 @@
...
+- Reduce risk of failure, such as by ensuring sufficient milestone granularity, feasability, and considering history of completion or non-completion by the parties involved.
+- Opportunity cost of allocating funding here versus elsewhere, and a baseline bar for the work being worth the funding.
+- Conflicts of interest among the funder, the proposal, the relevant governing parties, or the affected areas of work.
+
+The SC can also set criteria in advance that rule out cases such as specific categories, projects, funders, recipients, where the likely harm outweighs the benefit. These standing criteria are reviewed from time to time and can be added or lifted the same way. The SC will also revisit a standing criterion when the facts that motivated it have shifted substantially.
+
+SC members record and own their reasoning for every decision, alongside the SC's unified reasoning, and may approve or decline on any of these grounds. The SC refines the criteria as it learns, including when a funded effort ends up harming the ecosystem, so the decisions that follow are sharper.
+
+# Considerations
```
fricklerhandwerk, line 117: This document is missing measurable outcomes or success criteria, making it impossible to judge whether the stated goals were reached by the implementation.

proposals/grants-and-initiatives/README.md
fricklerhandwerk: That commit does not address my comment. What you get by a low flat review fee is likely no reviewers or useless reviews. What I meant by "price it into the proposal" is that a proposal should allocate the cost of reviews, which will be domain-specific and highly dependent on who's involved in the first place. Imagine: I propose to add a non-trivial feature to Nix, and maintainers get 5% for review. This is how you make nothingburgers.

Conversely, if any interesting work can only be done with maintainer involvement anyway, why not just scrap the public competition and give the money to maintainers directly to have them figure out on their own?

proposals/grants-and-initiatives/README.md
infinisil: Yes we know such individuals, it's often people with availability that want to improve certain aspects of the ecosystem, but cannot justify working on it for free against potential paid opportunities.

— @infinisil @Ra33it0

proposals/grants-and-initiatives/README.md
infinisil: The grant program is not to replace employment relationships. Grantees would be fully independent contractors and would not be employed by the foundation or potential funders.

See also https://github.com/NixOS/org/pull/270#discussion_r3880917859

— @infinisil @Ra33it0

proposals/grants-and-initiatives/README.md
infinisil: Because there's limited funding, and choices can only be made if you see the options, it doesn't work with continuous requests.

— @infinisil @Ra33it0

proposals/grants-and-initiatives/README.md
infinisil: This is the only absolute number in the document, but we've now adjusted it to scale with company revenue, so that smaller companies can still participate.

4ede6f3bff7bfb6ade215e13f88155514908fa62

proposals/grants-and-initiatives/README.md
SomeoneSerge: > Grantees would be fully independent contractors and would not be employed by the foundation or potential funders.

Precisely! That's what I mean by the "hidden employment relationship"! The grantees would be carrying out commercial work for the funders, directly contributing to the funders' profits, but they'd not get their share of the created margin (neither would the Foundation!), and they won't enjoy any of the so-called "security and benefits"

proposals/grants-and-initiatives/README.md
SomeoneSerge: You know of individuals who specifically want to channel a commercial customer's money through a non-profit organization that is legally barred from charging their customer a competitive price? :)

proposals/grants-and-initiatives/README.md
infinisil: Update: Will likely be revised some more

proposals/grants-and-initiatives/README.md
SomeoneSerge: > Because there's limited funding, and choices can only be made if you see the options, it doesn't work with continuous requests.

Choices can be made as long as you have a margin (e.g the EC and the German money that the Foundation has had). Given where Nix(pkgs) is now, the funding does not at all need to be limited, in fact it's first and foremost limited by our capacity to coordinate and spend. It's important the program doesn't accidentally limit the future margin and doesn't leave us with less choice than we enjoy now. This is why I hypothesize that the money should be reaching the Foundation through maintainers-operated entities in the form of a tax, and not from the for-profit partners with Foundation as a proxy

proposals/grants-and-initiatives/README.md
SomeoneSerge: Ty! There's also this diagram https://github.com/NixOS/org/blob/4ede6f3bff7bfb6ade215e13f88155514908fa62/proposals/grants-and-initiatives/initiatives.svg

proposals/grants-and-initiatives/README.md
KiaraGrouwstra: there may be a tension of interests here in 'security and benefits' to the developer (as a potential employee) vs the foundation getting the margins to address _their_ problems.
developers potential interested in taking this may then want to consider taking that into account - tho iiuc, that did already seem the norm.

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+# Detailed Design
+
+There are three inter-related parts to this proposal:
+
+- Team initiatives: A way for teams to request funding, which can come either from general-purpose donations or purpose-bound funding
+- Grant program: A way for contributors to propose ideas and get paid for delivering milestones. The funds can come from both general-purpose donations and purpose-bound funding.
+- Purpose-bound funding: A way for companies to provide funding for specific grants proposals or team initiatives.
+
+These programs should incorporate messaging to encourage contributing individuals to seek employment contracts or contracting agreements with funding partners.
```
jfly, line 23: I don't think I understand what "these programs" refers to. Is it the 3 items in the bulleted list above? If so, perhaps the preceding sentence ("There are three inter-related parts to this proposal") could be modified to make that clearer.

proposals/grants-and-initiatives/README.md
jfly: I'm also curious what the intent is here. Is this encouraging contributing individuals to bypass the nixos foundation?

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,126 @@
...
+([source](https://excalidraw.com/#json=rRutoi8nEcpjv1UIj3X87,VCPIeHZP_6ouG0U7tWQxMg), make sure to update the source link when changing anything)
+
+## Purpose-bound funding
+
+- Establish a process for going into contracts with companies to supply funding for specific purposes, with a guarantee that part of the funding will be used for the stated purpose, and that what could not be used as such is repurposed according to the funder companies wishes.
+  - Funding recommendations for differently sized companies should be given
+  - The [SC's evaluation criteria and process](#sc-evaluation-criteria-and-process) have to be met
+  - The board ensures standard legal requirements
+  - This can be used for [team-submitted initiatives](#team-initiatives) and grant proposals from the current feedback phase
+  - For team initiative funding, we charge an extra 10% for manegement
```
jfly, line 93: Nit:

```suggestion
  - For team initiative funding, we charge an extra 10% for management
```

jfly (commented):
@infinisil asked me to review this from my perspective as a member of the formatting and infra team. For the record, I have never sought or received a grant before. I have done some paid work for next packages arranged by nixcademy (the test driver container work).

Overall, this seems like a cool thing to try out! I see some concerns raised by @SomeoneSerge that this will result in people getting paid less than if they had sought out the work directly. I have two thoughts on that:

1. It seems to assume that people have the inclination to pound the pavement and find entities who want to pay them. For my volunteer work, I have no interest in spending my limited time hustling like that.
2. The paid test container work I did is an example of something I had no intention of working on, but was happy to do because it was an opportunity to make things better and learn. Perhaps this program could also be a place for stuff like that, where people who have ideas but not time make grant proposals, and other people who have skills and time actually do the work.

Other misc thoughts follow.

It wasn't clear to me what a grant proposal would even look like. Are they for a fixed dollar amount, or some rate with an expected range of hours? Maybe this would be clear to people who have done grant work before.

It might not belong in this doc, but I think it would be helpful to have some concrete examples of things this is supposed to enable. However, perhaps the best way to do that is to just launch the program and then people will see examples in real life :)

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,126 @@
...
+- Wave midpoint: Close proposal submissions, announce the feedback phase and reach out to:
+  - Various stakeholders for feedback, including authoritative community teams or individuals
+  - Companies for the opportunity to submit extra funding for specific proposals (via purpose-bound funding)
+    - Also mention team initiatives, same as on the website
+- Feedback phase (1.5 months): Collection of feedback and funds
+  - Feedback from stakeholders can be discussed and acted upon
+  - Sign purpose-bound funding contracts with funding companies wanting to submit funds for specific proposals
+  - Any proposal PRs submitted during this time will only be eligible for the next wave
+- Wave end: SC decides which proposals get funded, while prioritising the agenda and satisfying the [SC's evaluation criteria and process](#sc-evaluation-criteria-and-process), which are then merged to signal that the proposal is accepted for funding and the work can start
```
djacu, line 64: No time constraint on SC deliberation.

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,126 @@
...
+    - Proposals may also be for non-official projects
+    - Timelines can be arbitrarily long and have arbitrarily many milestones in them
+  - Teams and funding partners are encouraged to directly synchronise with each other about the scope of the work and grant amount. Ideally the team submits something to the grant program that the company then submits funding for using the purpose-bound funding mechanism
+- Wave midpoint: Close proposal submissions, announce the feedback phase and reach out to:
+  - Various stakeholders for feedback, including authoritative community teams or individuals
+  - Companies for the opportunity to submit extra funding for specific proposals (via purpose-bound funding)
+    - Also mention team initiatives, same as on the website
+- Feedback phase (1.5 months): Collection of feedback and funds
+  - Feedback from stakeholders can be discussed and acted upon
+  - Sign purpose-bound funding contracts with funding companies wanting to submit funds for specific proposals
```
djacu, line 62: Contracts are signed before the funding decision. What happens if the SC rejects?

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,126 @@
...
+- Wave end: SC decides which proposals get funded, while prioritising the agenda and satisfying the [SC's evaluation criteria and process](#sc-evaluation-criteria-and-process), which are then merged to signal that the proposal is accepted for funding and the work can start
+  - Proposals can get rejected due to either:
+    - No SC majority to back the proposal -> Proposal PR gets closed
+    - SC majority to back the proposal, but not enough funding this round -> Proposal PR is kept open, in case there's enough funding for a later wave
+      - Be clear that in such a case, there's still no guarantee to receive funding later
+  - The board ensures standard legal requirements
+
+### Milestone evaluation
+
+After grant proposals are kicked off, the applicants are expected to do the work and follow up after completing each milestone with detailed reports submitted as a PRs to the evaluation repo. An appointed evaluator will evaluate the work, request changes if necessary, and give approval if satisfactory, in which case the applicants will be paid.
```
djacu, line 73: What's the terminal state for a failed or abandoned grant?

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,126 @@
...
+To make the SC's funding decisions consistent, reviewable, and defensible, this sets out the criteria the SC evaluates against and how it records each one.
+
+In addition to the existing requirements, the SC weighs whether a funder or proposal could harm the project, and whether the work being funded is genuinely useful.
+
+Criteria under consideration include
+- Reputational, financial, or moral harm to the project.
+- Applicability and usefulness, whether consensus from relevant stakeholders exists or is needed, and whether the SC would be needed to shepherd the effort.
+- Reduce risk of failure, such as by ensuring sufficient milestone granularity, feasability, and considering history of completion or non-completion by the parties involved.
+- Opportunity cost of allocating funding here versus elsewhere, and a baseline bar for the work being worth the funding.
+- Conflicts of interest among the funder, the proposal, the relevant governing parties, or the affected areas of work.
```
djacu, line 117: I'd like to see written COI/recusal rules for SC, board, and evaluators

djacu (commented):
I was asked by @Ra33it0 to review this on behalf of the Marketing Team. Excited to see where this goes. Commented with some initial thoughts. I'll try to look more.

proposals/grants-and-initiatives/README.md
infinisil: We'll not specify it here, there's many more implementation details on how the SC makes this decision as well that will be needed for the concrete implementation

proposals/grants-and-initiatives/README.md
infinisil: The funds will be repurposed according to the funders wishes, see https://github.com/NixOS/org/blob/d6fd3a619b16df7b5cf22607fae18bf75f9f6933/proposals/grants-and-initiatives/README.md?plain=1#L87

proposals/grants-and-initiatives/README.md
SomeoneSerge: Hopefully it is, but we should not encourage employment

thilobillerbeck (approved):

proposals/grants-and-initiatives/README.md
roberth: Adding *why* would be nice, but I'd be ok with
```suggestion
## Goals
```

proposals/grants-and-initiatives/README.md
```diff
@@ -0,0 +1,125 @@
...
+- Evaluators together get 5% of each milestone amount for its evaluation, independent of whether the work is passing or not
+  - The 5% is budgeted internally by the foundation
+
+![](./waves.svg)
+
+([source](https://excalidraw.com/#json=rRutoi8nEcpjv1UIj3X87,VCPIeHZP_6ouG0U7tWQxMg), make sure to update the source link when changing anything)
+
+## Purpose-bound funding
+
+- Establish a process for going into contracts with companies to supply funding for specific purposes, with a guarantee that part of the funding will be used for the stated purpose, and that what could not be used as such is repurposed according to the funder companies wishes.
```
roberth, line 87: Entering into contract may have implications for ANBI status.
Surely the work product being publicly available helps? But the contract would seem to reframe it as a service *to* a specific entity. That seems (to me, IANAL) difficult to count towards ANBI's 90% common good rule.
I think there's multiple ways out of this apparent conflict.

a. Drop the ANBI ambition. Misses out on benefits of "true donors" so to speak, but certainly workable. Stichting Blender Foundation operates this way.

Without ANBI, Dutch donors face schenkbelasting and lose their giftenaftrek, and foreign donors and grant-makers lose the credential they may rely on for deductibility in their own jurisdiction. VAT has to be charged on contract work as usual.

b. Make a number of adjustments so that the funding is by donation. Donations can carry a stated purpose, take effect only once the SC has accepted the proposal, and return if unspent, but nowhere near the guarantees a commercial contract can provide. Revocation by the donor would be too controlling for ANBI.

This setup would be close to the Python Software Foundation's model.

ANBI feasible, provided the Stichting keeps the final say on which proposals are funded and who carries them out (in practice the SC, under a delegation from the board). No VAT on donations.

c. Move the contract work to a B.V. (private corp) owned by the Stichting. This would safeguard the ANBI status, but add some bureaucratic overhead, due to the extra entity and arm's-length contracts for Stichting resources the B.V. would get to use (e.g. establishing a "market rate" for trademark royalties).
This overhead may or may not be offset by the extra clarity it affords. For instance, a B.V.'s bookkeeping is routine work for any accountant, whereas accounting for an ANBI-compliant stichting is more specialised and probably worth keeping small.
In this setup the B.V. pays the Stichting arm's-length royalties (e.g. for the trademark) and fees for any Stichting resources it uses; those are costs for the B.V. The remaining margin is taxed as normal before flowing to the Stichting as dividends.

Housing the purpose bound funding in a B.V. also sends a signal to agencies that the Stichting's commercial arm competes with them.
While this might not be the intent, tax codes are set up to level the playing field, which I believe works out well. The arm's length rules make competition fair, and indeed it is in the best interest of the foundation to let anyone, including third party agencies contribute towards the purely functional deployment model in the ways that they are uniquely suited to. We'll want to highlight that.

A setup like this could also host non-development services that are perhaps not feasible to crowd-fund and should be paid by interested parties. (Some ideas, possibly quite bad, but their existence suggests good ones exist. Examples: prioritized builds of certain packages on cache.nixos.org, validation of reproducibility and security through independent rebuilds)

This is close to how Mozilla and Wikimedia operate; OpenSSL uses a sibling corporation rather than a subsidiary.

(Side note on Blender to avoid confusion: Blender has used a "working company" structure since 2007, where Blender Institute B.V. carries employment, office and other operational liabilities while IP and funds stay in the Stichting; they are currently completing the transfer of the B.V.'s ownership to the Foundation. That setup sits primarily on the expenditure side — donations still arrive at the Foundation — and it has served well to keep the project and its IP insulated from operational liabilities, but it is not a solution for the funding side of things. This is a common pattern ("werkmaatschappij") that could be applied for a three-entity setup, but that seems like overengineering at this point.)

d. Apache model, where companies pay the developers and the foundation pays for the other things. This fails to compile funders into FTE-scale bundles, which I believe to be an important goal.

roberth (changes requested):
I hadn't considered this in the context of ANBI before.
It would impact the grant program one way or another, so that ambition would best be taken into consideration now.
I've done a bit of research into paths that could be taken, as a starting point, but I am not a lawyer.

ctheune:
From an entity potentially providing funding: I'm happy to see this being worked on and sorry for not giving feedback earlier. From my perspective there's lots of small details that will have to be figured out as this grows and seeing how this works out properly, but I like the general setup. 

One aspect on the UI example that you showed: when providing funding I'd prefer to set my contribution, see the percentages taken away and then see what can be allocated to which parts. Entering some "random" amount that will then get added on will make it a game of trying to see how I can fit my budget in. The unknown taxes not withstanding ... (we should clarify what is to be expected/what happens on that side, too, though).

PR #275, opened 2026-07-23
values: specify deterministic automation

pbsds:
Two years ago few would think to include artificial intelligence under the umbrella term "automation", this is now changing. With this change my aim is to preserve the original spirit of our values as our language and environment shifts. If my interpretation of this value is flawed, please let me know.

### DIFF
--- a/doc/values.md
+++ b/doc/values.md
@@ -72,7 +72,7 @@ We build trust primarily by working together on Nix projects.
 
 We are a global community, and disseminating information and maintaining processes can be difficult.
 We are also a large project with a lot of hard and repetitive work.
-Therefore, we value automation over toil, while recognizing that not all toil can be automated.
+Therefore, we value deterministic automation over toil, while recognizing that not all toil can be automated.
 Automation reduces toil, but people are still accountable.
 Adding new toil needs a very strong justification.
 


roberth:
> as our language [...] shifts

Shift in environment is clear. By shift in language, you mean that the word automation has changed?
I personally don't feel like it has, but I don't keep up with the AI cheerleading crowd.

### Phrasing / form

```diff
-Therefore, we value automation over toil, while recognizing that not all toil can be automated.
+Therefore, we value deterministic automation over toil, while recognizing that not all toil can be automated.
```

The values should be centered around us humans, so a technical detail like "deterministic" doesn't really belong here on epistemological grounds.
Specifically, what you're describing seems to be a change of norms, not one of values.

To illustrate the difference, someone may respect all life equally and therefore choose veganism. The prior is a value, and the latter is a personal norm or rule.

As a concrete test, a formulation like "deterministic" technically brings into question the validity of useful tools like probabilistic [property checking](https://en.wikipedia.org/wiki/Software_testing#Property_testing) or probabilistic data algorithms like [RFC 195's](https://github.com/NixOS/rfcs/pull/195#issuecomment-4425411428). Those are useful and their nondeterminism harmless, but I feel this would need clarification, at which point it doesn't feel like a well-formed value anymore.

### Current values

Most applicable to your observation is our value that

> Automation reduces toil, but people are still accountable.

This line has absolutely grown in relevance. It is important to me, and I believe it is reflective of values that are widely held, including among those in the community who have a positive view on "AI".

### Logic

In terms of logic, your proposal is a weakening of a precondition, making it a smaller statement, which technically always fits more people into the group, so into the community. I like that.
However, language is fuzzier than logic, and a qualifier like that will read more strongly than logic prescribes, i.e. as a negative statement about that which is excluded.
I do not like that connotation.

### My opinion / verdict

Besides technical arguments about the validity and merits, I owe you my opinion as an SC member.

I believe maintainers and contributors should be free to choose their own tools, even if they could use them unproductively or to ruin their reputation, however unfortunate that is. If they didn't care about the quality of their contributions, I'm glad we can uncover that problem more easily now.
On the flip side, an increase in low quality drive-bys (ie without responsible humans) can become a real problem, but so far on a manageable scale that doesn't warrant crude measures that would be in line with the polarization that has played out in other projects, at least for the time being.

That said, I do agree deterministic tooling is better than the unprincipled and biased choices produced by "AI".

Nonetheless, I believe this edit to be a slight net negative, mostly on the basis of it moving away from being a value and more towards a norm or rule, and partly on the basis of the quoted line already representing my perspective on the issues.

tomberek:
Assuming the intent for this is to exclude AI/LLMs from the automation Value; i disagree because I think there is a space for such tooling in our domain. The question is when and under what conditions. But to say it has no value or excluding it from the Value seems short-sighted; applying today's evaluation of a technology to a long-lived Value.

I also don't think the proposed change would meet the intended outcome. We have deterministic LLMs. We have useful tooling that is non-deterministic as @roberth points out.

So for me, this fails on both fronts; both on the intent and outcome. And I'll +1 @roberth's post above.

pbsds:
I'd love to hear from the rest of the SC and community, but I respond early since it seems we read this value differently. I'm not trying to ban nor discourage people from using LLM tooling with this amendment. My read of this value is not about interactive/user tooling at all, but rather about CI/GHA, maintenance, and our formal procedures.

tomberek:
> I'd love to hear from the rest of the SC and community, but I respond early since it seems we read this value differently. I'm not trying to ban nor discourage people from using LLM tooling with this amendment. My read of this value is not about interactive/user tooling at all, but rather about CI/GHA, maintenance, and our formal procedures.

Thanks for the clarification; my assumption for your intent was wrong.

philiptaron:
Like Robert, I've also grown to believe that this line is quite important.

> Automation reduces toil, but people are still accountable.

My feeling is that determinism is only weakly correlated with automation that reduces toil, and that reducing toil is about the confluence of social, technical, and machine energy to move things from what a person cares about over and over and over again in a grinding way to something that a person does not.

A deterministic machine which is no longer aligned with the social needs of the community and mindlessly opens and closes issues, for instance, wouldn't be toil-reducing but toil-increasing.

A non-deterministic machine - or one for whom the question isn't categorically possible, like a code review bot - very well might reduce toil. The fact the current ones often don't doesn't eliminate the possibility, I feel.

I think that rounds down to "I don't think a values update is needed here". Though as always I appreciate you and your thoughtfulness Peder. ❤️

pbsds:
I agree with you that this edit does not accurately capture the nuance that I intended. We could workshop "deterministic" for many other words, but as you pointed out: "people are still accountable" already does the job quite well. I thank you for your input, it's been a week now and I feel this can be closed.

PR #288, opened 2026-09-18
Fix CI

infinisil:
Getting this in CI runs, like
https://github.com/NixOS/org/actions/runs/31025754088/job/92373643252?pr=276:

> Error: Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.


infinisil:
Maybe not necessary anymore, https://github.com/NixOS/org/pull/287 merged without problems

